PR #5, opened 2017-04-15
Update icons to proper brand colors

9mm:
* Snapchat isn't correct but that requires a black stroke around the ghost to be accurate


edent:
Thanks for these. I've manually added most of them (in upper-case).

It looks like some brands have multiple style guides - so there's not always a definitive answer.

Really appreciate you taking the time to contribute to this. The beer's on me :-)

9mm:
Nice, although I recommend updating Reddit to orange, which is their primary color:

https://about.reddit.com/

(see their logo, etc)

9mm:
Also, for the "Phone" icon, it seems to be modeling something similar to the iPhone call icon, so I picked Apple's official green color for their call button

https://developer.apple.com/ios/human-interface-guidelines/visual-design/color/

It's not exactly the same icon in iOS though so take it or leave it. The current green just seems quite garish.

edent:
Good shout - thanks!

PR #16, opened 2017-04-16
youtube-tiny.svg had a missing svg namespace

RazrFalcon:


edent:
As per the README, that's to show just how small you can make it. The file renders in most apps, but not on the web.

RazrFalcon:
My bad.

PR #20, opened 2017-04-16
11% smaller stackoverflow icon

RazrFalcon:


RazrFalcon:
It's because of 100% -> 512 conversion.

Issue #1, opened 2017-04-15
Website/API


franciscop:
I repeat my comment here from HN:

This is really great. As someone who has done quite a few size-sensitive projects/experiments, have you considered providing them in a small website? So artists can contribute with more icons and developers who want to use them can modify them when downloading (border radius, background, etc). I could throw a quick prototype if that sounds good.

edent:
Go ahead :-)

There are already lots of Icon / SVG sites. I got many of the original vectors from https://www.iconfinder.com/ - but I do like the idea of being able to change the border radius and background programatically.

I'm not sure about having people submit designs. User Generated Content is very hard moderate.

franciscop:
That's really the main idea that I wanted for making it, to allow for user generated content to be uploaded. What would you do if people did a PR here with new logos?

However with a search I see that https://www.iconfinder.com/search/?q=facebook&price=free is really similar for instance (615 bytes one of them), so yeah it might not make sense to do that...

If we remove that functionality, it'd be only to modify the background+radius, which *could* (but maybe it shouldn't) just be statically generated and hosted in github. It'd be 51kb per logo folder approx: 1kb/logo * 50 (1-50%) + 1kb

franciscop:
I have made a small preview site:

![radius_0](https://cloud.githubusercontent.com/assets/2801252/25078853/5c4cfb9a-2337-11e7-8683-95b7f04eaf0a.png)

![radius_5](https://cloud.githubusercontent.com/assets/2801252/25078851/5b8c9e18-2337-11e7-97a0-c22ca5961710.png)

![radius_8](https://cloud.githubusercontent.com/assets/2801252/25078852/5c2b72cc-2337-11e7-939a-d17338415be3.png)

![radius_32](https://cloud.githubusercontent.com/assets/2801252/25078854/5c88faf0-2337-11e7-8aef-2b851311acc8.png)

PR #23, opened 2017-04-17
Small tool for radius visualization

franciscop:
Added small website tool for radius visualization following #1


franciscop:
The thing is that they were not consistent at all so they fir better (IMHO) this way and Instagram also has a logo that is white over colorful background.

But sure, will revert it, HTML5 and Mastodon are already quite different.

PR #26, opened 2017-04-18
52% smaller VK icon with small fix

RazrFalcon:
It looks exactly the same except top right curve. I've fixed it a bit. But I can revert it back. I will not change size much.


edent:
Gorgeous!

PR #30, opened 2017-04-18
Optimised PNGs

Martin-Pitt:
Ran original PNGs with ImageOptim first for lossless compression. (ImageOptim is nice and automated)

Then created a new set by going through ImageAlpha for more serious cost-savings. (ImageAlpha is great when you want to get serious)


Martin-Pitt:
Thankfully the SVG is superior as it's an awesome format. However wanted to give PNG a bit more of a fighting chance.

I do wonder if the margins are less when the logo becomes more complex.


To go off on a bit of a tangent:-

If I wanted to be REALLY nitpicky...

Obviously the PNGs really hurt from the 512x512 resolution, which is not exactly a common size I see icons at.
If I take the github.png and resize it at 32x32 and scale it up to account for a DPI of 3.0 (96x96) and optimise it with ImageAlpha it sits at 710b, which is less than the svg equivalent of 765b.

HOWEVER, I think SVGs have so many advantages that should be highlighted in this repo beyond cost saving, such as being able to include it directly in a HTML page and take advantage of features such as being able to recolour it with CSS dynamically or in context, animate it, or do whatever with it.
Also the ability to re-use an asset easily across projects or in any context without having to worry about whatever resolution it is at and that changing the colour is one quick change in your favourite code editor.

PR #32, opened 2017-04-19
33% smaller mastodon icon

RazrFalcon:
1. Unlike other icons it uses href linking. Is it ok?
2. It's a bit different from original, but more symmetrically correct. Maybe we should send them this icon.


edent:
1. I don't know :-) is href well supported?
2. Hahah! Good plan!

RazrFalcon:
1. Yes. It's a basic SVG feature.

edent:
Well alrighty then :-)

PR #33, opened 2017-04-19
35% smaller wikipedia icon

RazrFalcon:
It's a new icon based on an official wiki logo. Because it's look different I'm using pull request instead of direct push so you can approve this changes.


edent:
I think this is the official one - https://commons.wikimedia.org/wiki/File:Wikipedia%27s_W.svg

Looks more like the original to me.

RazrFalcon:
I've used this: https://commons.wikimedia.org/wiki/File:Wikipedia-W-bold-in-square.svg

edent:
Interesting. They seem to use the thin version for their mobile apps. 

https://play.google.com/store/apps/details?id=org.wikipedia&hl=en_GB

https://en.m.wikipedia.org/wiki/List_of_Wikipedia_mobile_applications

Let's keep the styling thin for now.

RazrFalcon:
Looks like this image should be used as a reference: https://upload.wikimedia.org/wikipedia/commons/b/b3/Wikipedia-logo-v2-en.svg

edent:
I'm going to China for a week, so won't be able to access GitHub.

Issue #43, opened 2017-05-13
Shrink Inner Icons


joezimjs:
If you bump the radius to 50% to make the icons circular, some of the icons (linked in, RSS, maybe more) extend beyond the colored circle background.

edent:
Interesting. The icon box is supposed to be 400*400 in a 512*512 wrapper. So what's the maths for the optimal size?

RazrFalcon:
It's a huge change. A lot of icons can't be fit in the circle, so they should be changed/rescaled, which includes making them pixel-perfect again and this takes a lot of time.

joezimjs:
400x400 makes the diagonal dimension 565px, which is larger than the 512px diameter of the circle. If you drop the icon down to 350x350, the diagonal will be 495px which should work well. (Pythagorean theorem for all this stuff) Wouldn't you be able to multiply all dimensions/coordinates by 7/8 to make this work? I might give it a try myself when I get to my PC.

edent:
Would love to see them!

joezimjs:
Nevermind. I forgot how complicated SVG paths are. It'd take me a while just to learn enough to understand how they work. :stuck_out_tongue_closed_eyes:

A potential alternative would be to increase the viewbox and rectangle size to 600 and just reposition everything. Which I actually know enough to do (I think). :smile: How do you guys feel about that?

edent:
Try it. Send a PR and we'll see what it looks like. 

Might be easier just to alter the ones which break the bounds of the circle though.

joezimjs:
Submitted PR #44

edent:
Now tracked in #86

PR #44, opened 2017-05-15
Enlarged the viewbox and background rectangle of many icons; Fix for #43

joezimjs:
... so that the icons would fit nicely inside the background if rx was set to 50%.

Some went to 640 instead of just 600 so that they look even less cramped, especially when the icons are small.


joezimjs:
The biggest reason I shrunk the others is because when you shrink the icons (48x48 for example), you may not see the border, or it may only be 1px or so at that size.  Hackernews and StackOverflow are sitting right against the edge. You seem to be ok with them though, so I'll undo the changes to those icons.

PR #45, opened 2017-05-31
Removed newlines from SVGs

AceLewis:
New lines are not needed in SVGs so I removed them.


edent:
Thanks for this Ace, I appreciate the work. As it says in the readme "Newlines can be stripped - they've been kept for readability where possible."

Newlines also make it easier for seeing diffs in pull requests.

I'm going to keep the newlines (for now) - but if you come up with any other space saving measures, I'd love to see them.

PR #62, opened 2017-11-27
Run SVGO

CapacitorSet:
Although the README says you ran the images through SVGO, I tried doing that myself and it gave better results (sometimes far better, eg. with the Windows logo at -30%). Either way, I figured I'd merge the updated version.


CapacitorSet:
I see.

I ran `svgo -f tiny -p 2 --pretty --multipass` (SVGO v1.0.3), and it still shows some improvements over the original files (as well as some losses due to --pretty). Also, since dropping `--pretty` shows an improvement, I added the compressed versions in `tiniest/`.

edent:
This is brilliant! Two minor things, though.

1. `role="img"` needs to be present. I think SVGO has stripped it out. That's required for accessibility.
2. A few of the icons have updated since you've done your PR. Would you mind making sure yours are based off the new ones.

Once again - thank you so much for your contributions. They're extremely cool!

CapacitorSet:
Okay, should be alright now. Since the setup was becoming more complex, I added SVGO config files in `tiny/` and `tiniest/` for ease of use.

edent:
OMG! You are amazing. 10 Points to Ravenclaw!

PR #70, opened 2017-11-27
Badoo

jrfferreira:
Adding dating app Badoo's logo


edent:
Many thanks for your contribution!

PR #76, opened 2017-11-27
Fix Google icon

### LINKED ISSUE #57: Incorrect Google icon
I've never seen that Google logo used as a square icon. Yes, it's the G in the full Google logo, but the Google logo used for things like Google's Oauth is similar to the favicon, having four colours in the G. Perhaps there few reasons to use the logo in the first place as there isn't much you can link on Google that doesn't have its own icons, like Blogger, Google Maps, or Google+. But even if my website linked to Google Books or Google Scholar, I would probably use one similar to the official favicon used for Google Search and Google Sign-In.

Evolving Google Identity

_"The Google G is a compact version of the Google logo that works in small contexts."_  
 https://design.google/library/evolving-google-identity/

The official download for the Google G for use in the icon can be found [here](https://madeby.google.com/static/images/google_g_logo.svg).

ErikHumphrey:
Fixes #57

473 or 770 bytes. May have removed too many tags.


edent:
Thanks both! BTW @ErikHumphrey - I spent a year abroad studying at Carleton. Go Senators!

PR #79, opened 2017-11-28
RSS icon changed to a single path

baybal:


baybal:
Added accidentally removed aria label

baybal:
I was checking with high res original. The original Twitter logo has no Beziers as such, all curves in it are supposed to be arcs of only 3 radii as per their design guideline. Why I think it is important? Beziers will inevitably loose precision when you trunkcate the decimals and can only emulate an arc fragment if one side of a Bezier mirrors another. While they can come visually close, calculus-wise they aren't. Twitter logo is remarkable for being mathematically determenistic and analytically complete.

baybal:
@edent what do you think ^?

Issue #69, opened 2017-11-27
Add captions 


mqudsi:
Especially as the number of icons grows, adding captions/labels to the chart would be helpful to find the logo one is looking for via ctrl+f or similar.

edent:
That's a good plan - please can you send a Pull Request to fix it?

mqudsi:
Sorry, not invested in the project.

PR #82, opened 2017-11-28
Add captions for easy searching

florisluiten:
As said in issue #69, captions allow for searching a specific icon. I've implemented the captions in the README.md as well as in the radius.html. The latter uses the aria-label of the SVG, so icon builders do not need to do anything special for that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,113 +17,113 @@ Originally created for my contact page - [https://edent.tel/](https://edent.tel/
 
 <table>
 <tr>
-<td><img src="https://edent.github.io/SuperTinyIcons/images/svg/hackernews.svg" width="125" title="Hacker News" /><br>228 Bytes</td>
-<td><img src="https://edent.github.io/SuperTinyIcons/images/svg/flickr.svg" width="125" title="flickr" /><br>252 Bytes</td>
-<td><img src="https://edent.github.io/SuperTinyIcons/images/svg/facebook.svg" width="125" title="Facebook" /><br>270 Bytes</td>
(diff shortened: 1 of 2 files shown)


edent:
Wow! That's fantastic. Thank you so much 😀

florisluiten:
Dude, you're quick. 

You're welcome, glad to help

edent:
There's a bug on the radius tool - it keeps adding the label multiple times. Happening on Firefox on Android, haven't checked other browsers yet.

florisluiten:
Ah yes, you are right. I'm fixing that right away

edent:
Sorry, had to revert this for two reasons.

1. On mobile, it lead to inconsistent icon sizes on the main page.
2. The radius tool was still adding extra titles.

I think it is a good idea - just need to work out how to do it properly.

florisluiten:
You are right, we have to do it correctly.

1. I don't fully understand what you mean. Are you talking about the radius.html or README.md?
2. I seem to have made a mistake in this PR. The other PR #83 was merged and should add the label only once on radius.html. That seems to work fine.

edent:
Here's a screenshot of the README.  As you can see, the length of the caption influences the size of the table cell.  This means some icons display smaller than others.
screenshot_20171129-133337

florisluiten:
I see now what you mean. It looks like I'm not able to fix this, since GitHub does not render the "<colgroup>" I've added to my version of the markdown file. It looks like we have to live with the fact that we cannot display the labels in the markdown file and keep the layout nice. I think it's more important to display the icons in the same size, so no labels in the README.md file then.

I guess that's okay though, since the radius.html file is linked from the markdown file, which does contain the labels.

PR #87, opened 2017-11-29
33% smaller mastodon icon

surajrimal:
1. Unlike other icons it uses href linking. Is it ok?
2. It's a bit different from original, but more symmetrically correct. Maybe we should send them this icon.


olivesayang:
33% smaller mastodon icon

edent:
Hi @surajrimal this looks great! Would you mind using SVGO to shrink the file and then committing it to master?

edent:
Looks like mastodon have changed logo - https://github.com/tootsuite/mastodon/blob/master/public/apple-touch-icon.png

PR #91, opened 2017-11-30
Guidelines & More Ubuntu Optimizations.

arturoacevedob:
Added guidelines with instructions.

Ubuntu keeps getting smaller, now about 400 bits less than original version.


arturoacevedob:
You know what. It solves a very, very niche scenario. I think it's better to take out and don't cause confusion. I'll keep a copy just in case.

edent:
Fab! Thanks! I'm going to move the Guidelines into a different folder, and incorporate your instructions into the README.

arturoacevedob:
Perfect.

PR #94, opened 2017-12-01
Uber And Google Play

tonyraoul:
Add Uber Logo and Google Play logos
Uber has many other logos but this one is the most minimal


arturoacevedob:
Those are great. Let's add some things:

1. In your opening `<svg>` tag, add `viewBox="0 0 512 512"` to set the display size.
2. In the same place, add `aria-label="Uber or Google Play" role="img"` for accessibility reasons.
3. For [radius.html](https://edent.github.io/SuperTinyIcons/radius.html) to work, we use a `<rect>` with the `rx="15%"` rounded corners property as a default background. So you can replace `<path fill-rule="evenodd" d="M-1.304 1.644h512v512h-512z"/>` with `<rect width="512" height="512" rx="15%"/>`

After that, I think Edent will approve and merge.

edent:
Fixed in #97

PR #97, opened 2017-12-01
Gmail and Google Play Logos

tonyraoul:
add gmail and google play logos


edent:
Thanks Tony!  The GMail one looks broken in the preview - any way of fixing that?

gmail

arturoacevedob:
@tonyraoul How did you fix the Chrome issue?

tonyraoul:
it is all about floating points , I was to aggressive removing floating points I just restored few of them until it is back in shape

arturoacevedob:
Thanks for the response!

Issue #93, opened 2017-11-30
BitBucket has a new logo


edent:
https://www.atlassian.com/company/news/press-kit

PR #99, opened 2017-12-01
Aligned to guideline and path optimized Android and Bitbucket

arturoacevedob:
All were aligned accordingly and optimized to the following percentages:

Android - 31%
Bitbucket + 3%

Updated README with the new file sizes.


edent:
Fixes #93

PR #101, opened 2017-12-02
Corrects vendor colors for Evernote & GitLab

TheLonelyGhost:
The Evernote icon didn't look quite right with a white background, so I took an eyedropper tool to the Evernote.com site and grabbed the elephant icon's exact hex color. Changed the fill color in the SVG to match that.

Similar story with GitLab, but since the original logo is so multicolored, I opted for the main color [their site](https://about.gitlab.com) uses for emphasis. It visually looks close enough to the original logo and it better mimics their branding, which is the whole point here, right?

Final change is to correct an issue with the Meetup reference logo. Looks like someone decided to rename a PNG and thought that would convert it to a JPG file, then upload it to the Akamai CDN. Running `file images/reference/meetup.jpeg` from the terminal indicated it was a PNG originally, so I made the change to the file extension and the `index.md` file where it was referenced.


edent:
Nifty!

PR #114, opened 2018-01-13
added XING

fileformat:


edent:
Thanks for this! I think it can be even smaller though.

1. The first line should be `<svg viewBox="0 0 512 512" xmlns="http://www.w3.org/2000/svg" aria-label="Xing" role="img">` - you don't need the XML declaration at the top of yours.
2. Try running your file through [SVGO](https://github.com/svg/svgo) or similar - that will cut the size down.

fileformat:
Done

edent:
Thanks!

PR #119, opened 2018-01-18
Fix JavaScript syntax error in `radius.html`

### LINKED ISSUE #118: JavaScript syntax error in radius.html
The [radius example](https://edent.github.io/SuperTinyIcons/radius.html) doesn't work due to a syntax error in the embedded JavaScript: https://github.com/edent/SuperTinyIcons/blob/master/radius.html#L83

jdanford:
Fixes #118

### DIFF
--- a/radius.html
+++ b/radius.html
@@ -34,7 +34,7 @@
 			opacity: 0.25;
 		}
 		.underlay {
-			position: relative; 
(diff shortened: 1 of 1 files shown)


edent:
Thanks for this :-)

PR #120, opened 2018-01-27
added pinboard

fileformat:


edent:
Thanks for this :-)

For future reference, for accessibility, you need to add `aria-label="Pinboard" role="img"` to the `<svg`.  You can also do the corner radius by just using `rx="15%"`.  I'll also drop the precision to zero decimal places, which will save some bytes.

I'll add a note to the README to make it clearer in future.

fileformat:
Can you share the recommended svgo command line (or config)?  I'm using them in for a side project ([SimpleShare.IO](https://simpleshare.io)) and have a couple more in the works.

edent:
Sure! We use https://github.com/edent/SuperTinyIcons/blob/master/images/svg/svgo.yml

I might move that to a more obvious place. Of course, nothing beats a little bit of hand tweaking ;-)

Love your new project BTW. Can't wait to use it.

PR #127, opened 2018-01-30
Achieved 9% size reduction

baybal:
Used a large arc, quadratic curves instead of cubic ones, and a repeat instructions


edent:
I'm granting you commit access to the Repo.  This is per the advice of [The Pull Request Hack](http://felixge.de/2013/03/11/the-pull-request-hack.html).

PR #129, opened 2018-02-04
Buffer

fileformat:


edent:
Thanks! Again, related to #130 - why does this get a coloured background when the official logo doesn't?

fileformat:
I make [several versions](http://www.vectorlogo.zone/logos/buffer/index.html) of each logo, but my "icon" version has a clear background and goes to the edges, so you would need to tweak it to match your template.

baybal:
@fileformat here is a smaller version of your geometry

aria-label="Buffer" role="img"
viewBox="0 0 512 512"><rect
width="512" height="512"
rx="15%"
fill="#000"/><path id="a" d="M84 251q-7 5 0 10l162 75q10 3 20 0l162-75q8-5 0-10l-32-14q-10-4-21 0l-109 50q-10 5-20 0l-109-50q-10-4-21 0z"/><use width="100%" height="100%" transform="translate(0 94)" xlink:href="#a"/><path d="M84 166q-7-4 0-9l162-75q10-4 20 0l162 75q8 5 0 9l-162 75q-10 5-20 0z"/></svg>

edent:
Thanks for this. @fileformat I'm granting you commit access to the Repo.  This is per the advice of [The Pull Request Hack](http://felixge.de/2013/03/11/the-pull-request-hack.html).

PR #134, opened 2018-02-11
Add dribbble icon

thomasbritton:


edent:
Hi Thomas. That's a good start. Needs a bit more work to get it consistent with the other icons. Let me know if you need any help. Looking forward to having this one!

PR #153, opened 2018-05-21
Use self-closing tags

Vimux:
* Sketch 528 -> 492 bytes
* Slack 852 -> 804 bytes
* workato 495 -> 489 bytes


edent:
Awesome work!

PR #165, opened 2019-01-26
Create Dribbble icon and make associated edits

judeGibbons:


edent:
Looks great! I reckon you can get it even smaller if you drop the precision down. I see lots of decimal places 😀

PR #166, opened 2019-01-26
Create W3C icon and make associated edits

judeGibbons:


edent:
Icon looks good. Two things before I can merge...

1. There's a conflict with  `images/reference/index.md` and `radius.html` - they don't include the changes you made in the Dribble PR.
2. Can you make the icon smaller? Using something like https://jakearchibald.github.io/svgomg/ you can drop the precision, which should help.

Thanks :-)

judeGibbons:
OK, will do! I'm manually editing them but hadn't sussed out about the precision when I did those. Should I submit each new icon in a separate branch (which will always cause a conflict in the radius file)? And if I want to tweak a few at once should I also submit them separately or can they all go in one PR?

edent:
Best to do them separately, I think it might be time to get rid of the Radius file.

judeGibbons:
Could logos be added to the radius file on separate lines? It could be reformatted every now and then to prevent it getting too long.

Issue #170, opened 2019-02-19
Losslessly compress all SVGs


fulldecent:
The SVGs can be losslessly compressed. This means that the identical image can be expressed with less bytes. 

<img width="798" alt="screen shot 2019-02-18 at 10 38 48 pm" src="https://user-images.githubusercontent.com/382183/52988617-f9934000-33cd-11e9-9841-d979fd5be260.png">

Try using the tool ImageOptim.app to do it.

PR #171, opened 2019-02-19
compress svgs

fulldecent:


fulldecent:
fixes #170

baybal:
I will also appreciate is somebody can make handedited gziped versions of icons.

edent:
Nope, sorry. Removing accessibility information is *not* lossless.

The reference images are straight copies of the original files and do not need to be shrunk.

If you can find genuine space savings on the generated SVGs, that would be great. But removing accessibility information is not up for debate.

fulldecent:
Is the `aria-label="ICON NAME" role="img"` the only semantic information in the files?

Some files currently have this and some do not.

* Example without: https://raw.githubusercontent.com/edent/SuperTinyIcons/9886063af83fc2cb9d296d0918e3d49295aefbff/images/reference/codepen.svg
* Example with: https://raw.githubusercontent.com/edent/SuperTinyIcons/master/images/svg/vimeo.svg

The names seem to be in the readme, and the mapping of A.svg -> B is found with:

```sh
sed -ne 's|.*svg/\(.*\).svg".*title="\(.*\)".*|\1,\2|p' README.md
```

If yes, then we have a path to minify the assets which keeps required accessibility attributes. Perhaps we will keep existing files in /svg-inkscape/ (since they are actually Inkscape-flavored files), other options are /svg-editable/ and /svg-inkscape-editable/. And the minified images with accessibility can be output to /svg/. 

Please confirm if such an effort would be in scope to be merged for this project.

edent:
Thank you so much for your efforts. Let me explain the structure more clearly. The README I've written obviously isn't clear enough.

* `images/svg/` contains all the optimised SVGs. These are designed to be as small as possible.
* `images/reference/` contains reference images. These have been copied from the official sources. They do not need to be optimised. They are just so that you can compare them to the optimised version.

So, please do not modify anything in `images/reference`.

As for semantic information, there are two things.

You are correct that `aria-label="ICON NAME" role="img"` is there for accessibility. Again, this must not be removed.

Secondly, we use vertical whitespace to make it easier for humans to understand the file, and to see changes in it.  A space and a newline take up the same number of bytes.  So changing:

```
aria-label="Access" role="img"
viewBox="0 0 512 512"><rect
```
into
```
<svg xmlns="http://www.w3.org/2000/svg" aria-label="Access" role="img" viewBox="0 0 512 512"><rect
```
Does **not** save any space.

If you can make savings in `/images/svg/` without removing the accessibility data, or the whitespace, then I would be very happy to accept pull requests.

PR #185, opened 2019-03-15
Update duckduckgo

judeGibbons:
Reduce DuckDuckGo logo to 1020 bytes including the xmlns:xlink attribute.


edent:
Fab! You should have the power to merge your own code. If not, let me know and I'll give you another invite.

PR #186, opened 2019-03-15
Add Linux logo

judeGibbons:


edent:
I'm sorry, but this is WITCHCRAFT! I was stumped by the Tux logo and could never get it small enough. BRILLIANT work!

judeGibbons:
Thank you very much! 😊

PR #187, opened 2019-03-15
Add unicode logo

judeGibbons:


judeGibbons:
I see what you mean - my laptop screen is a bit small! Yes, I'll take a look at that. By the way, I linked the Unicode logo to their logo policy but that is technically in contravention of their logo policy...

edent:
Lovely!

PR #188, opened 2019-03-16
Add NHS icon

mattedgar:
Referencing NHS Identity Guidelines: "The NHS logo is protected by law. It is a UK trade mark owned by the Secretary of State for Health and Social Care.  It is also protected by copyright. Only original artwork files for the NHS logo should be used. You should not attempt to recreate it yourself."


edent:
Brilliant! Thanks. I've squooshed it a little more.

Issue #131, opened 2018-02-10
Missing Apple icons and existing icon problems


MarkSmit:
New icons:
* Name: Apple Music
* URL: [Apple Music website](https://apple.com/apple-music)
* Link to icon: [Apple Music icon](https://is2-ssl.mzstatic.com/image/thumb/Purple118/v4/d2/cf/1e/d2cf1e56-a429-c4cb-e973-c53f01176bdb/AppIcon-1x_U007emarketing-0-0-GLES2_U002c0-512MB-sRGB-0-0-0-85-220-0-0-0-9.png/738x0w.jpg)
-----
* Name: Apple Podcasts
* URL: [Apple Podcasts website](https://www.apple.com/itunes/podcasts/)
* Link to icon: [Apple Podcasts icon](https://is3-ssl.mzstatic.com/image/thumb/Purple128/v4/bb/45/69/bb45692f-2d72-9d98-f802-918d0f560624/mzl.liqqlbrn.png/738x0w.jpg)
-----
* Name: App Store
* URL: [App Store website](appstore.com)
* Link to icon: [App Store icon](https://images.apple.com/v/ios/app-store/b/images/overview/app_store_icon_large_2x.jpg)
-----
* Name: Apple Maps
* URL: [Apple Maps website](maps.apple.com)
* Link to icon: [Apple Maps icon](https://is1-ssl.mzstatic.com/image/thumb/Purple128/v4/1d/12/10/1d12104a-ce4e-68e0-e8b2-76c9487f8890/mzl.vecqidkx.png/738x0w.jpg)
-----
* Name: Apple Homekit
* URL: [Apple Homekit website](apple.com/homekit)
* Link to icon: [Apple Homekit icon](https://is4-ssl.mzstatic.com/image/thumb/Purple71/v4/1c/86/3c/1c863c3e-c93b-548a-1eb0-09460284b10f/mzl.hpkipaqm.png/738x0w.jpg)

These are very useful for when you want to share your content on one of these platforms, like music playlists or locations in maps.

Some things about existing icons:
* The instagram icon doens have the gradient as shown on the [real icon](https://is5-ssl.mzstatic.com/image/thumb/Purple118/v4/e6/95/bf/e695bfa6-9e66-625f-c7a4-6365ac3fd366/Prod-1x_U007emarketing-85-220-0-5.png/738x0w.jpg)
[...]

PR #190, opened 2019-03-19
Update apple music icon

judeGibbons:
Added stop-opacity as suggested in Issue #131; this works on Safari 8.


edent:
Working in Safari. I'll merge.

PR #209, opened 2019-06-01
Add safari logo

judeGibbons:


baybal:
I suggest using single path with markers for dashes

judeGibbons:
> I suggest using single path with markers for dashes
I didn't try it that way, but I figured that drawing the vertical and horizontal dashes and then rotating would make the endpoints of the dashes exact whereas marking each endpoint would need at least 1 decimal point each time to make them look accurate. I may be wrong though!

judeGibbons:
I've done a quick test and converting the ring of marker lines to a single compound path more than doubles the size of that part compared to using the rotated version, and that's with 0 decimal places which looks wrong. Using 1 decimal place looks ok but the size of that path alone is 1022 bytes!

PR #215, opened 2019-07-23
Optimize PNG images, and update size comparison in README

mvasilkov:
Hi, I've optimized the PNG files using [ImageOptim][1] and updated their sizes in the README.

I feel this leads to a more fair size comparison, since SVG files were also optimized using svgo + svgcleaner. The SVG icons are still smaller by a factor of 3–5.

[1]: https://imageoptim.com/mac


edent:
Nice

PR #216, opened 2019-07-23
Add Julia programming language

cormullion:
I'll be very pleased (and a bit surprised) if I've made no mistakes adding a Julia icon.


images/svg/julia.svg
```diff
@@ -0,0 +1,12 @@
+<svg
+xmlns="http://www.w3.org/2000/svg"
+aria-label="..." role="img"
```
edent: You need to add some text in the aria label.

images/svg/julia.svg
```diff
@@ -0,0 +1,12 @@
...
+viewBox="0 0 512 512">
+<rect
+width="512" height="512"
+rx="15%"
+fill="#fff"/>
+<circle fill="#389826" cx="256" cy="137" r="83"/>
+<circle fill="#cb3c33" cx="145" cy="329" r="83"/>
```
edent: You can remove the newlines to save a few bytes.

images/svg/julia.svg
cormullion: Ha, I thought that might be some SVG/XML wizardry... :)

images/svg/julia.svg
cormullion: 🏌️‍♂️

Issue #217, opened 2019-07-23
Confusion about gradients


granthusbands:
The front readme says icons must have "No gradients or fades.", but the Firefox and JSON logos clearly have gradients in them. Should that rule be removed, as the size requirement will preclude many gradients, anyway?

edent:
Yup. Go ahead and remove it.

PR #218, opened 2019-07-25
Removed gradient/fade rule

granthusbands:
Firefox and JSON both use gradients. The file size limit keeps things simple enough. Discussed in #217.


edent:
Great work!

PR #221, opened 2019-07-28
Add OpenBenches icon

judeGibbons:


edent:
I *love* it!

judeGibbons:
I did try a version where the logo filled the whole 512x512 square, and changing the radius cropped it down to a circle keeping the black outline. I didn't think it complied with the branding though!

PR #227, opened 2019-08-14
Add Threema Icon

medienverbinder:
Adding the Threema icon based on the iOS icon from the press material.


edent:
Looks brilliant! Thanks :-)

Issue #229, opened 2019-09-19
Add Orcid


edent:
## The icon I want is:

* Name: Orcid
* URL: https://orcid.org/
* Link to icon: https://github.com/ORCID/ORCID-Source/blob/master/orcid-web/src/main/webapp/static/img/id-icon.svg

* Vector version of the logo: https://github.com/ORCID/ORCID-Source/blob/master/orcid-web/src/main/webapp/static/img/id-icon.svg
* Official style guide or brand guide:

PR #233, opened 2019-09-30
Added Orcid Icon

agungjsp:
- [x] New file
- [x] Edit Readme
- [x] Add reference image
- [x] Edit reference page
- [x] (optional) Android Image

#229


edent:
Thank you so much for this. I think the icon needs a little more work. At the moment, it seems squished in the corner:
Screenshot_2019-09-30 Added Orcid Icon by agungjsp · Pull Request #233 · edent SuperTinyIcons

agungjsp:
I just committed, is everything fine?

edent:
It is *very* close! Take a look at some of the other icons. We use this template:

```
<svg
xmlns="http://www.w3.org/2000/svg"
aria-label="..." role="img"
viewBox="0 0 512 512"><rect
width="512" height="512"
rx="15%"
fill="#fff"/> ...
</svg>
```

So, if you can fix the viewbox to be the same as the others, I can accept it :-)

agungjsp:
I'm sorry, the SVG won't work properly when I use the template, so SVG will automatically be like this:



If I change the ViewBox and width to ```256```, it might works.

Issue #138, opened 2018-02-27
Printer Icon


leocaseiro:
## The icon I want is:

* Name: printer
* URL: https://www.iconfinder.com/icons/326675/print_icon
* Link to icon: https://www.iconfinder.com/icons/326675/print_icon#size=512

* Vector version of the logo: https://www.iconfinder.com/icons/326675/print_icon#size=512
* Official style guide or brand guide: any

edent:
Fixed in #235

PR #235, opened 2019-10-01
Add print/printer icon #138

piriys:
Refer to: #138 
- [x] New file (print.svg)
- [x] Edit Readme (in Misc section)
- [x] Add reference image
- [x] Edit reference page


edent:
That looks great! Thank you so much. I think you can probably squeeze a few more bytes out of it by trimming the whitespace.

PR #240, opened 2019-10-04
fb added to optimized imgs

tanaynayak:


edent:
This appears to be the old logo?

PR #242, opened 2019-10-04
WIP: many icons

romain-dartigues:
Hi, I pre-submit my PR for feedback before it is finished; any comment would be appreciated (like some icons are not welcomes or if you prefer one PR per icon).

## Checklist

- [x] New file
- [x] Edit Readme
- [ ] Add reference image
- [ ] Edit reference page
- [ ] (optional) Android Image


edent:
Those are *brilliant*! Thank you so much. I'd prefer one PR per icon - or group of icons - just because it makes the diffs a bit easier for me. Also means if one icon isn't quite right, it won't hold up the others.

romain-dartigues:
Thank you for your encouraging answer; I started with #242 and am just waiting for a follow up from you to know if I do the same for the other logos (and more).

romain-dartigues:
Replaced by #243, #272, #273, #274, #275, #276, #277.

PR #243, opened 2019-10-04
add: mattermost

romain-dartigues:
Using official EPS template from 2016 downloaded from https://www.mattermost.org/.


edent:
Excellent stuff! I reckon I can shave off a few bytes by removing some newlines - but this is a great first PR. Thank you!

PR #248, opened 2019-10-05
Add git icon

hirakJS:


edent:
Could you please make this fit in with the other icons? It should fit within a circle.  You can also shrink the size by simplifying the transform.

hirakJS:
done changes.

hirakJS:
Do I need to add the icon into README file?

edent:
Fixed in #272

PR #249, opened 2019-10-05
Added flutter icon

leanthonyml:


edent:
Great stuff! I've added this - but if you remove some of the whitespace, I think you can get it even smaller :-)

leanthonyml:
On it! 😄

PR #272, opened 2019-10-07
add: git SCM

romain-dartigues:
Logo from https://git-scm.com/


romain-dartigues:
Note : somehow duplicate with #248.

PR #276, opened 2019-10-07
add: RedHat ansible

romain-dartigues:
Ansible logo, black version.


edent:
This needs a background rectangle, I think. Either plain white, or extend the background circle.

Issue #265, opened 2019-10-07
add Protonmail


alexanderadam:
## The icon I want is:

* Name: Protonmail
* URL: https://protonmail.com
* Link to icon: https://protonmail.com/images/media/logos/protonmail-sign-purple.eps
* Official style guide or brand guide: [protonmail.com/media-kit](https://protonmail.com/media-kit#Logos)

alexanderadam:
Maybe it's a but too late But it seems there's still some problem with the Protonmail icon (besides the missing `</tr><tr>`):

![Screenshot_20191014_111027](https://user-images.githubusercontent.com/372620/66740353-405d0b00-ee73-11e9-9519-c44e42e84597.png)

PR #278, opened 2019-10-07
Add Protonmail

zreese:
ref #265


alexanderadam:
It's mindblowing how fast you are!
Awesome, thank you so much! :pray:

images/svg/protonmail.svg
```diff
@@ -0,0 +1,9 @@
+<svg
+xmlns="http://www.w3.org/2000/svg"
+aria-label="Protonmail"
+viewBox="0 0 512 512">
```
romain-dartigues: You could save over 20 bytes by removing the spaces, the group and putting the `fill=...` into the `<svg` tag.
Also, I think it would require the white background rectangle.

images/svg/protonmail.svg
```diff
@@ -0,0 +1,9 @@
...
+  <g fill="#8A90C7">
+    <path d="M259 77s-108-6-131 114v79s0 12 23 29c28 17 91 68 108 68s74-51 102-69c23-16 23-28 23-28v-79C361 71 259 77 259 77zm68 159H185v-45c11-57 74-57 74-57s57 0 68 57v45z"/>
+    <path d="M259 384l-34-12-97-68v120s0 11 17 11h222c11 0 17-11 17-11V304l-97 69-28 11z"/>
+  </g>
+</svg>
```
edent: You can remove these newlines to shrink the SVG even further.
Please can you do this as a purple background with white foreground, like the reference image?

edent:
Yup - if you can make those small changes, we can merge it. Thanks!

images/svg/protonmail.svg
zreese: Is it a white background rectangle with a color circle (rounded rect) on top? In that case, I can knock the paths out of the circle if that ends up being smaller.

PR #288, opened 2019-10-09
Add Deezer Logo

matheusagcosta:
Closes #281

Issue #281, opened 2019-10-08
Add Deezer podcast logo


edent:
## The icon I want is:

* Name: Deezer 
* URL: https://www.deezer.com
* Link to icon: https://deezerbrand.com/d/9wmMErzrRuiH/brand-elements

* Vector version of the logo:
* Official style guide or brand guide: https://developers.deezer.com/guidelines/logo

lihofm:
I will take care of this :)

PR #301, opened 2019-10-13
add: deezer

romain-dartigues:
Note that the Deezer brandbook explicitly forbid using the equalizer without the wordmark text, except when the company name will be written below.

Unfortunately I was unable to keep it under 1KiB without sacrificing something, even with xlinks and patterns...

Credits to @matheusagcosta challenging me to shrink it under 1024 B.

Relates to #281, #288.


edent:
Magic!

PR #309, opened 2019-10-14
Fix broken src attribute of Protonmail image

chesio:
This has been missed in #278.


alexanderadam:
Could you please also [add the missing `</tr><tr>` so that it's six elements per line](https://github.com/edent/SuperTinyIcons/issues/265#issuecomment-541568500)?

edent:
Thanks!

PR #310, opened 2019-10-14
Add Gitea icon

### LINKED ISSUE #256: add Gitea
## The icon I want is:

* Name: Gitea
* URL: https://gitea.io/
* Link to icon: https://github.com/go-gitea/gitea/blob/master/public/img/gitea-512.png
* Vector version of the logo: https://github.com/go-gitea/gitea/blob/master/assets/logo.svg

zreese:
Fixes #256


edent:
A good attempt - but doesn't look quite right to me.

Ideally should be centred.
gittea

Could you also remove the drone stuff from this PR. That'll make it easier to merge.

edent:
I reckon you can make it even smaller by removing the first path. It also contains some stuff like `stroke-miterlimit="15.1"` which don't seem to add anything.

I'll merge it now - but look forward to an even tinier version :-)

PR #314, opened 2019-10-14
addition of conformed xmpp svg

wuxmedia:
hopefully


wuxmedia:
ah jeez I don't want to merge all this crap :D

PR #320, opened 2019-10-16
Remove outline path from Gitea icon

zreese:


edent:
Brilliant! Please can you update the size in the README and I'll merge.

PR #321, opened 2019-10-16
Added Ubiquiti Networks logo

vangyyy:


edent:
Looks good, but please can you follow the template at https://github.com/edent/SuperTinyIcons/blob/master/CONTRIBUTING.md

We need the aria label for accessibility. And the layout helps with diffs.

```
<svg
xmlns="http://www.w3.org/2000/svg"
aria-label="..." role="img"
viewBox="0 0 512 512"><rect
width="512" height="512"
rx="15%"
fill="#fff"/>...</svg>
```

vangyyy:
Will fix that, also for #324.

edent:
Fab!

PR #322, opened 2019-10-16
add ruby on rails logo

fenwick67:
I couldn't find exact brand guidelines, nor a large-scale square logo, so I used the favicon as the reference.


alexanderadam:
Awesome work!

edent:
That's a perfect first commit. Thanks!

PR #300, opened 2019-10-13
Regenerated Android vector drawables

vangyyy:
Recreated all existing android vector drawables and added missing ones that already existed in SVG formats within repository. Fixes #299. I used [this](https://svg2vector.com/) tool, which eliminates problems introduced when converting with [current tool](https://inloop.github.io/svg2android/). It would be a good practice to require this procedure for all newly created icons in the future.

PR #324, opened 2019-10-16
Fixed luckyframework alignment and smoothed edges (also half the size)

vangyyy:
Much smaller icon
Smoother and aligned edges


vangyyy:
Android drawable already fixed in #300.

edent:
Think you've included the wrong file in this commit. The SVG looks like an Inkscape export.

edent:
Thanks. Just to note, I've added a space into the ARIA label. That's so screen readers can pronounce it properly. I think there are a few already in the repo I need to correct.

vangyyy:
Yes, I looked at one and it didn't have space, so I didn't use one also. When I find some missing spaces in the future, I'll fix them.

PR #337, opened 2019-10-21
Added Facebook Messenger svg and android drawable

jmb:


jmb:
I think this has done it (after a lot of battling with the tools and manually working out the gradient!). For reference, I also found this site which was helpful: https://lean-svg.netlify.com/

edent:
Good stuff! You'll notice that I remove some newlines to shrink the file even more.

jmb:
> Good stuff! You'll notice that I remove some newlines to shrink the file even more.

Thanks! Out of interest, why do you leave the new lines at the top?

PR #339, opened 2019-10-22
debian swirl

wuxmedia:
Not the most fantastically faithful to the original icon. I think balance of resolution and size is good though. 
Thoughts?


alexanderadam:
I like this compromise :+1:

wuxmedia:
I think I've resolved the readme correctly.

wuxmedia:
I'm not sure - the Gimp download on their site is D70751 (which is my preferred colour)
but as you note the SVG is A80030, which is the reference given I suppose. (we are so literally painting the bikeshed here :D )

jmb:
How odd! Have checked what the jpg/png version is? I wonder which it is supposed to be!

wuxmedia:
the JPEG is... hideous. Depends on what pixel you sample :D 
deb
their PNG is cleaner but #C70036 ¯\_(ツ)_/¯

wuxmedia:
ahh ok - I think I'll follow the official SVG.

edent:
I think this one might be too complicated to get down under 1,024 bytes.

positive-tdh:
Is that a 'nope' to the smoothed down version? we have 400 bytes to play with. Might look at adding some nodes  (i was pretty brutal at chopping it down, but further tweaks  (thanks @jmb) made it a lot smaller.

edent:
Yeah, I think restoring some nodes would help. It looks too different at the moment.

wuxmedia:
fair point. closing this pull

PR #340, opened 2019-10-23
add: HAML, fix #307

### LINKED ISSUE #307: add HAML
## The icon I want is:

* Name: HAML
* URL: http://haml.info/
* Link to icon: https://github.com/haml/haml/issues/819#issuecomment-66153259 / http://cl.ly/0H0B3V2B3G3Q

romain-dartigues:
The logo has been aggressively simplified to fit under 1KiB,
it look more like the favicon than the full quality one.


images/svg/haml.svg
```diff
@@ -0,0 +1,7 @@
+<svg xmlns="http://www.w3.org/2000/svg"
+aria-label="HAML" role="img"
+viewBox="-11 -11 512 512"><rect
+width="490" height="490"
+rx="15%"
+stroke="#29291c" stroke-width="22"
```
edent: I'm not sure about the stroke around the edge. Unless that's part of the original design?
Rest looks pretty good though!

romain-dartigues: > I'm not sure about the stroke around the edge. Unless that's part of the original design?

How about now?
If you consider their [favicon](http://haml.info/images/favicon.ico) to be their "original design", yes it was (but barely visible at it's size).

edent:
This is great! I've just noticed that some renderers require "xmlns:x=..." to be `xmlns:xlink..."

That'll bump up the file sizes of a few, but I'll take care of that.

PR #343, opened 2019-10-24
add CentOS

romain-dartigues:
After RedHat, CentOS!

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,6 +249,7 @@ Say thanks!
 </tr>
 <tr>
 <td>RedHat<br><img src="https://edent.github.io/SuperTinyIcons/images/svg/redhat.svg" width="125" title="RedHat"/><br>566 Bytes</td>
+<td>CentOS<br><img src="https://edent.github.io/SuperTinyIcons/images/svg/centos.svg" width="125" title="CentOS"/><br>762 Bytes</td>
 <td>Git<br><img src="https://edent.github.io/SuperTinyIcons/images/svg/git.svg" width="125" title="git"/><br>482 Bytes</td>
 <td>Microsoft<br><img src="https://edent.github.io/SuperTinyIcons/images/svg/microsoft.svg" width="125" title="Microsoft"/><br>435 Bytes</td>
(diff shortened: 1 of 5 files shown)


edent:
Very clever work! Doesn't seem to render in Firefox for Android. Let me check what's causing the issue.

edent:
Aha! Two small things, and one irrelevant thing.

1. `use href` doesn't work on some renderers unless you explicitly use `xmlns:xlink="http://www.w3.org/1999/xlink"`
2. The `style` stuff *should* work - but doesn't seem to affect the width on all renderers.
3. No need for `<g id="h">` - just added weight.

Here's a new version at 762 bytes. Only slightly larger than yours, but should work everywhere.

```
xmlns:xlink="http://www.w3.org/1999/xlink"
aria-label="CentOS" role="img"
viewBox="0 0 512 512"><rect
width="512" height="512"
rx="15%"
fill="#fff"/><g id="g"><rect x="118" y="118" fill="#9ccd2a" width="124" height="124"/><rect x="270" y="118" fill="#932279" width="124" height="124"/><rect x="270" y="270" fill="#efa724" width="124" height="124"/><rect x="118" y="270" fill="#262577" width="124" height="124"/></g><use xlink:href="#g" transform="rotate(225 256 256)"/><path d="M118 270h124v124H118zm152 0h124v124H270zm0-152h124v124H270zm-152 0h124v124H118zm245 226l-87-88 87-88 88 88zM256 236l-88-87 88-88 88 88zM149 344l-88-88 88-88 87 88zm107 107l-88-88 88-87 88 87z" fill="none" stroke="#fff" stroke-width="9"/></svg>
```

romain-dartigues:
> 1. `use href` doesn't work on some renderers unless you explicitly use `xmlns:xlink="http://www.w3.org/1999/xlink"`

Well shoot, and I just though I was smart finding [xlink was deprecated](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/xlink:href)... :,^(

> 2. The `style` stuff _should_ work - but doesn't seem to affect the width on all renderers.

Dang, here goes my optimization once again.

> 3. No need for `<g id="h">` - just added weight.

Oh, you are right, a remnant of my trials and errors.

> Here's a new version at 762 bytes. Only slightly larger than yours, but should work everywhere.

Thanks for your time :^)

PR #350, opened 2019-10-28
Tweak Gradle icon

jmb:
Change style="fill:#02303a" to fill="#02303a" and reduce size a little. Makes more consistent with other icons too.


edent:
Good work - thanks!

PR #357, opened 2019-10-31
Ruby add

wuxmedia:
does this work? I can't see my svg, I suppose as it's reference to this repo :roll_eyes:

PR #358, opened 2019-10-31
Update index.md

wuxmedia:
is this right?


romain-dartigues:
I think you should add this commit to your previous MR #357.

Issue #348, opened 2019-10-26
Fix Rust


edent:
The Rust icon has a problem.

Some rendering engines require XLINK to be in the form `xmlns:xlink="http://www.w3.org/1999/xlink"`

The Rust icon uses `xmlns:x="http://www.w3.org/1999/xlink"`

If we convert to proper XLINK syntax, the files is 1,053 bytes - that's too large.

Otherwise, it looks like this:
![Screenshot at 2019-10-26 17-43-56](https://user-images.githubusercontent.com/837136/67623021-3abed800-f818-11e9-8443-69b494cb6ec5.png)

romain-dartigues:
Well, it look like a crowned R :^D.
I prefer fixing the redering engine than the Rust logo, a wonderful job was made on this icon...

More seriously, it would be nice to have an informative checklist of sofwares/versions the logo *must* render correctly (ie: chrome 70+, firefox 64+, firefox mobile 68+, etc.).

For example, software using `libpixbufloader-svg` (geeqie, nemo, ...) fail miserably to render andOTP, elementaryos, gojek, grafana, haml, keybase, redhat, rust, sentry, tutanota, wekan.

judeGibbons:
I remember this issue and there was a decision make to move from using xlink: to x: ([here](https://github.com/edent/SuperTinyIcons/issues/184#issuecomment-473293345)) but I see that later the problems were flagged with doing it this way. Happy to look at redoing mine, if I can...

PR #359, opened 2019-11-06
WIP: Minimize !@#%

romain-dartigues:
Minimize some more (not to be merged as-it).

I started hunting bytes, and I'd like a feedback first.
Am I doing right? Still want one MR per icon or could I propose a batch (with one icon per commit)?

* Lucky: remove unused objects, use a gradient instead of three shapes; not supported by svg2android
* XMPP: optimize path, use cloning and transform, unsupported by svg2android
* Intel: optimize path and remove the ®
* Ubiquiti: merge path, remove unused objects


images/svg/xmpp.svg
```diff
@@ -1,6 +1,7 @@
 <svg xmlns="http://www.w3.org/2000/svg"
+xmlns:x="http://www.w3.org/1999/xlink"
```
edent: This needs to be `xmlns:xlink=...` otherwise some rendering engines get confused.

edent:
It's addictive, isn't it :-)

Happy for you to do them in batches as you seem to know what you're doing.  I've left a note about the XLINK issue. Other than that, they look ready to go.

images/svg/xmpp.svg
romain-dartigues: Oh, I took it from Rust (yeah, I know, #348), but ok.

edent:
These all look good. Merge whenever you think they're ready.

edent:
Happy if you want to merge these.

edent:
As it has been more than a year, I'm going to cherry pick some of the logos from here which still need doing.
Sorry for the delay. It has been a strange year...!

PR #360, opened 2019-11-07
Bigger microsoft edge (new logo Nov 2019)

codepo8:
Here is the SVG version of the new logo, I guess you're going to make it smaller? :)


edent:
Cheers matey!

edent:
Fixed in #378 378

PR #364, opened 2019-11-10
add: Linux Mint

romain-dartigues:
There have been a discussion in linuxmint/linuxmint#78 for a new, simplified, logo; in case you want it, you can [see it there](https://github.com/romain-dartigues/SuperTinyIcons/blob/add-linux_mint-y/images/svg/linux_mint.svg) (366B).

Because of the shape I choose to not enclose it in a rounded square, what do you think?

Also, to keep it under 1KiB, I had to use the controversial XML entity hack I learned by looking at the Apple music icon.


romain-dartigues:
Logo changed, see: https://www.linuxmint.com/rel_tricia_cinnamon_whatsnew.php#artwork

PR #384, opened 2020-04-18
Added SVG

17jiangz1:


edent:
This is a *brilliant* first submission! Thank you so much. I will accept it - and then I think I can shrink it a little further.

Good work!

PR #394, opened 2020-04-21
Update svg.svg

17jiangz1:
Cleaned up and smallified


edent:
Excellent! Now that it is smaller, can we make it a bit more like https://www.w3.org/Graphics/SVG/svglogo.svg ? For example, the grey shape behind the letters?

17jiangz1:
The official guidelines https://www.w3.org/2009/08/svg-logos.html do not have any shading of the stylised version

PR #403, opened 2020-04-21
Update ubisoft.svg

17jiangz1:
Resized to specs/minified


edent:
Looking at their website, this should be white on a black background.

17jiangz1:
You're probably right. I was likely referring to their favicon, but white on black seems to be more oft used.

Issue #371, opened 2019-12-13
Add Uplay


jhasse:
## The icon I want is:

* Name: Uplay
* URL: https://uplay.ubisoft.com
* Link to icon: https://icon-library.net/icon/uplay-icon-8.html

* Vector version of the logo: I haven't found one yet.
* Official style guide or brand guide: Don't know if that even exists.

The24thDS:
Is this issue resolved?

edent:
It is, good spot!

PR #404, opened 2020-04-21
Create uplay.svg

17jiangz1:
Ref: https://commons.wikimedia.org/wiki/File:Uplay_Logo.png
Re: #371


edent:
Looks good. Please can you add to index.md and add a reference image to https://github.com/edent/SuperTinyIcons/blob/master/images/reference/index.md

PR #405, opened 2020-04-21
Update guideline.svg

17jiangz1:
Converted to actual circles (for usability), rounded colours


edent:
cx and cy should be at 256
Screenshot_2020-04-27 edent SuperTinyIcons

17jiangz1:
My bad!

edent:
I forgive you :-)

PR #406, opened 2020-04-21
Update rss.svg

17jiangz1:
Optimised -a few bytes

### DIFF
--- a/images/svg/rss.svg
+++ b/images/svg/rss.svg
@@ -3,4 +3,4 @@ aria-label="RSS" role="img"
 viewBox="0 0 512 512"><rect
 width="512" height="512"
 rx="15%"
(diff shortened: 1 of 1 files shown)


edent:
The circle needs to be filled.

17jiangz1:
The circle appears to be filled in Chrome, Opera, and Safari. It seems to be a quirk with Firefox... The same issue is present in Instagram's flash circle, albeit less visible.
image

17jiangz1:
Seems to be a bug in Firefox:
https://bugzilla.mozilla.org/show_bug.cgi?id=896248

edent:
Yes, I think the easiest thing to do is move the circle outside of the group. Then fill it.

PR #424, opened 2020-07-10
WIP: adding glitch icon

thisisjofrank:
I've made an SVG glitch icon (because i couldn't find a non-broken SVG glitch icon) but I don't know how to minimise it further and so can't get it below 5ish K, (nor do I know if that is possible!)

PR #425, opened 2020-07-10
Create glitch.svg

17jiangz1:
Re #424, colour could use some work


edent:
Really good. Take a look at #426 for some ideas on colour.

Issue #184, opened 2019-03-13
Fate of xlink, opinion needed


baybal:
XML preamble `xlink:` and namespace declaration `xmlns:xlink="http://www.w3.org/1999/xlink" `is no longer needed in every web renderer these days. https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/href

This has been a proposal since SVG 1.2, and this is why some renders not explicitly supporting 2.0 has its support.

Instead of `<use x="-19" y="170" xlink:href="#o"/>` you can simply use `<use x="-19" y="170" href="#o"/>`

Support in browsers is certainly there, but I can't say so for offline renderers. Librsvg doesn't seem to do it despite having most complete 1.2 support of all offline renderers.

edent:
Hmmm.... It looks like it still doesn't work in Safari. And Ubuntu doesn't seem to natively support it - either in EOM or file thumbnails.

I think we need to bring it back, for better compatibility. Or will that take any tiny SVG over the limit?

baybal:
Most Linux viewers use Librsvg. The question is if we continue this as a web first iconset.

Question about Safari, can you find somebody with a macintosh computer to check how things are in real life with last Safari version?

edent:
Safari Version 12.0.3 (13606.4.5.3.1) - these icons are broken
![Screen Shot 2019-03-14 at 14 16 14](https://user-images.githubusercontent.com/837136/54364000-fb19f600-4663-11e9-9797-d8919d6a53e1.png)
![Screen Shot 2019-03-14 at 14 16 20](https://user-images.githubusercontent.com/837136/54364001-fb19f600-4663-11e9-8e27-4ad4db4ec98f.png)
![Screen Shot 2019-03-14 at 14 16 34](https://user-images.githubusercontent.com/837136/54364007-fd7c5000-4663-11e9-9d3e-4c3cea1ffb77.png)
![Screen Shot 2019-03-14 at 14 16 43](https://user-images.githubusercontent.com/837136/54364008-fd7c5000-4663-11e9-9c1f-d9f119ca2a3f.png)
![Screen Shot 2019-03-14 at 14 17 00](https://user-images.githubusercontent.com/837136/54364010-fd7c5000-4663-11e9-9edc-e55bd9c84a43.png)
![Screen Shot 2019-03-14 at 14 17 12](https://user-images.githubusercontent.com/837136/54364011-fe14e680-4663-11e9-9495-48c47a3db2b5.png)
[...]

PR #426, opened 2020-07-13
Add Glitch website icon

judeGibbons:


edent:
This is AMAZING! The top fin is slightly wonky, but the rest is spot on.

17jiangz1:
@judeGibbons That looks much better! I've brought it down to 950 bytes [with a few minor changes](https://github.com/judeGibbons/SuperTinyIcons/pull/22), and was wondering if you could possibly figure out how to fix the wonky jagged gill line transitions/smooth out the top with the extra space?

judeGibbons:
The extra space would definitely allow for a better path, but I'm not sure about removing the xlink - as I understand it it's required in some situations, and for backwards compatibiity (e.g. it needs it in Safari 8) - see issue #184. I'm not sure at what point we decide it's no longer needed. I'll try tweaking the fin bit anyway.

PR #428, opened 2020-07-18
Make Minecraft icon cross-browser compatible - fixes #417

### LINKED ISSUE #417: Minecraft icon doesn't display in Safari WebKit engine
minecraft.svg displays as just a black box under Safari (both desktop and iOS)

judeGibbons:
This works across browsers but is large than the existing one (just scrapes in at 1023!).


edent:
You are magical!

Issue #429, opened 2020-07-28
Add Roundcube


alexanderadam:
## The icon I want is:

* Name: Roundcube
* URL: https://roundcube.net/
* Link to icon: https://roundcube.net/images/roundcube_logo_icon.svg

edent:
Hey Hacktoberfest participants! If you'd like to tackle this issue, [please follow the simple guidelines](https://github.com/edent/SuperTinyIcons/blob/master/CONTRIBUTING.md).

PR #431, opened 2020-10-04
Issue #429 - Add Roundcube Icon

ronan696:
#429 Added Roundcube icon, updated the readme and reference image


edent:
That's brilliant! Thank you so much for contributing. Feel free to add more icons.

Issue #430, opened 2020-07-28
Add Jellyfin


alexanderadam:
* Name: Jellyfin
* URL: https://jellyfin.org/
* Link to icon: https://jellyfin.org/images/banner-dark.svg

edent:
Hey Hacktoberfest participants! If you'd like to tackle this issue, [please follow the simple guidelines](https://github.com/edent/SuperTinyIcons/blob/master/CONTRIBUTING.md).

PR #432, opened 2020-10-04
Added Jellyfin logo, readme entry and reference

The24thDS:
This should complete #430


edent:
Hey! This looks really good. Could you please make a few tweaks and I'll merge the PR.

* Remove the `preserveAspectRatio="xMidYMid meet" ` it isn't needed.
* Remove the `xmlns:v="https://vecta.io/nano"` it isn't needed.
* Add a white background.
* Add an Aria Label.

So, the start of the file should look like this (exact formatting):
```
aria-label="Jellyfin" role="img"
xmlns:xlink="http://www.w3.org/1999/xlink"
viewBox="0 0 512 512"><rect
width="512" height="512"
rx="15%"
fill="#fff"/>
```

The24thDS:
Done :)

PR #436, opened 2020-10-05
Optimize Preact Icon

brunocerecetto:
* change White for #fff in svg file
* change readme with new size.


edent:
That's a great start. Take a look at the path. It has numbers in it like `221.7025033688164`

I don't think you need that level of precision. How does it look if you reduce it to an integer?

brunocerecetto:
It's no difference when i change these numbers to integer (in the image) and the size got smaller!

edent:
SVG is *magic*! Thanks for your contributions.

Issue #437, opened 2020-10-05
Add Gandi logo


edent:
## The icon I want is:

* Name: Gandi
* URL: https://www.gandi.net/en-GB
* Link to icon: https://www.gandi.net/assets/img/favicon/favicon_hires-ecc1d24b.png

* Vector version of the logo: https://news.gandi.net/wp-content/themes/news-gandi/assets/images/logo_gandi.svg
* Official style guide or brand guide: https://v4.gandi.net/static/images/illustrations/press/Gandi_LogoGuidelines_EN.pdf

Hey Hacktoberfest participants! If you'd like to tackle this issue, [please follow the simple guidelines](https://github.com/edent/SuperTinyIcons/blob/master/CONTRIBUTING.md).

PR #439, opened 2020-10-06
Add icon for Gandi

admvx:
Hello again! This one addresses #437

- [x] New file which is *under* 1,024 bytes
- [x] Edit Readme
- [x] Add reference image
- [x] Edit reference page
- [x] Android Image


edent:
Woo! Nice one.

PR #442, opened 2020-10-08
Add microformats logo

btrem:


edent:
Fantastic! Thanks 😁

btrem:
That was fast! The folks at microformats.org will (hopefully) be pleased.  (:

Issue #374, opened 2019-12-20
add KeePassDX


alexanderadam:
## The icon I want is:

* Name: KeePassDX
* URL: https://www.keepassdx.com/ / https://github.com/Kunzisoft/KeePassDX
* Link to icon: https://github.com/Kunzisoft/KeePassDX/blob/master/art/ic_launcher.svg

edent:
Hey Hacktoberfest participants! If you'd like to tackle this issue, [please follow the simple guidelines](https://github.com/edent/SuperTinyIcons/blob/master/CONTRIBUTING.md).

PR #443, opened 2020-10-08
Add KeePassDX icon

admvx:
Trickier one! 😅 Addresses #374 

- [x] New file which is *under* 1,024 bytes
- [x] Edit Readme
- [x] Add reference image
- [x] Edit reference page
- [x] Android Image


edent:
Amazing use of `feDropShadow`! I don't think we've had one of those before.

I'm granting you commit access to the Repo.  This is per the advice of [The Pull Request Hack](https://felixge.de/2013/03/11/the-pull-request-hack.html).

Please use this privileged wisely.

admvx:
Thank you @edent! I'll absolutely take the responsibility seriously – and I admire the philosophy of the linked post.

I must admit I had no cause to use `feDropShadow` before either 😁 Thanks again

Issue #362, opened 2019-11-08
Add cryptocurrency-related icons


samthelion17:
## The icon I want is: Blockchain Wallet

* Name: Blockchain Wallet
* URL: https://www.blockchain.com/
* Link to icon: https://cdn.freebiesupply.com/logos/large/2x/blockchain-logo-svg-vector.svg

* Vector version of the logo: https://cdn.freebiesupply.com/logos/large/2x/blockchain-logo-svg-vector.svg
* Official style guide or brand guide: 

## The icon I want is: Coinpot

* Name: Coinpot
* URL: https://coinpot.co
* Link to icon: http://www.bitcoins4free.ml/wp-content/uploads/2017/12/logo-1.png

* Vector version of the logo:
* Official style guide or brand guide: 

## The icon I want is: Humble Bundle

* Name: Humble Bundle
* URL: https://www.humblebundle.com/
* Link to icon: 

* Vector version of the logo: https://cdn.worldvectorlogo.com/logos/humblebundle.svg
* Official style guide or brand guide:

edent:
Hey Hacktoberfest participants! If you'd like to tackle this issue, [please follow the simple guidelines](https://github.com/edent/SuperTinyIcons/blob/master/CONTRIBUTING.md).

Sharps:
I thought I'd give this a shot for the hacktoberfest thing, made the blockchain wallet icon and then noticed that their logo is now different, heh. Still ended up with 557 bytes even though it looks very simple. No idea how you could fit the mess that is the coinpot logo into 1024 bytes.

I had fun though, when it comes to vectors for me everything would end with clicking save on illustrator, never really looked into the code before.

In any case, could you check if the one I made stands up to your standards ? Thanks
https://www.dropbox.com/s/s9m9rqciizqwc7p/blockchain.svg?dl=0

edent:
@Sharps that looks really good. Could you make one for the updated wallet? Make sure it follows the template in the CONTRIBUTING file and I'll happily accept a pull request.

PR #444, opened 2020-10-08
Humble Bundle icon

The24thDS:
This should take care of a part of #362


edent:
That's *really* close, but not quite right. If you set `rx="50%"` you'll see that the red stick goes outside the bounds. 
A broken image
Take a look at the guideline SVG in the Contributing file - that should help.

The24thDS:
Now it should be fine :)

admvx:
Thanks for the update @The24thDS :) I think the centering is better now, but the icon is too large with respect to the guide SVG:

image

Plus one naming issue from the contributing doc:
>Filename should be `nameofservice.svg` - all in lower-case. If a special character must be used, please replace it with an underscore (not a dot nor a dash).

The24thDS:
You are right. I will fix these as soon as I have some free time.

The24thDS:
Can you check it now @edent?

edent:
Lovely stuff. Well done!

PR #446, opened 2020-10-08
Save some more bytes

tobiasthaden:


edent:
Excellent work! Danke schön :-)

PR #447, opened 2020-10-10
Makes WhatsApp icon even smaller

btrem:
Uses a simpler path for the speech bubble, with
the white border defined by the stroke attribute,
and the inside green part done a standard fill.

Also changes the fill color to match WhatsApp’s
brand guide.

Final size: current version: 589 bytes
My update: 484 bytes


edent:
This is really good. Thanks!

Issue #420, opened 2020-06-03
add Samsung


alexanderadam:
this is probably a difficult one because I didn't found a proper circular one. But maybe the "S" could be used for that (like in [Samsung's favicon](https://cdn.samsung.com/etc/designs/smg/global/imgs/favicon.ico)? Otherwise [their app icon is the type in the Samsung blue](https://www.samsung.com/etc/designs/smg/global/imgs/app_ico.png).

## The icon I want is:
* Name: Samsung
* URL: https://www.samsung.com/
* Link to icon: https://en.wikipedia.org/wiki/File:Samsung_Logo.svg

edent:
Hey Hacktoberfest participants! If you'd like to tackle this issue, [please follow the simple guidelines](https://github.com/edent/SuperTinyIcons/blob/master/CONTRIBUTING.md).

logiclrd:
I'll give this one a go.

PR #454, opened 2020-10-20
Add Samsung logo (SVG)

logiclrd:
This PR adds 3 new icons:

* `samsung_s.svg`: This is the 'S' of the Samsung logo in Samsung blue filling a rounded-corner white icon.
* `samsung.svg`: This is the full Samsung name in white in a circle of Samsung blue. This icon file is 1023 bytes with LF line endings. Git informed me that it was going to change them to CRLF when committing, so it might now be slightly over. I spent a lot of time tightening this design by hand and making trade-offs to get more concise approximations. I don't know if there is much opportunity to further reduce the size.
* `samsung_swoop.svg`: This file does not quite meet the 1024-byte guideline. Either it is an okay exception, or it gets the axe. I don't think it is possible to make it significantly smaller in any way. This icon consists of the Samsung name in white on the tilted "swoop" ellipse in Samsung blue.

It is very difficult to fit the Samsung logo into 1024 uncompressed bytes. :-)


logiclrd:
Resolves #420.

images/svg/samsung_swoop.svg
```diff
@@ -0,0 +1,6 @@
+<svg xmlns="http://www.w3.org/2000/svg"
+aria-label="Samsung" role="img"
+viewBox="0 0 512 512" fill="#117cad"><rect
+width="512" height="512"
+rx="15%"
+fill="#fff"/><path d="M470 219c1 22-21 34-38 44-59 33-126 47-191 57-54 7-111 11-164-2-14-4-30-11-34-25-1-20 19-32 34-42 53-31 115-46 175-56 59-9 120-15 179-2 15 4 35 10 39 27z" fill="#034ea2"/><path fill="#fff" d="m118,232c-12 0-17 4-16 15 2 9 19 14 20 17l0 3c0 1.5-1 3-4 3-3 0-4.5-2-4.5-4l0-4-12 0c-1 12 8 16 16 16 10.5 0 16-3 16-13 1-14-18-15-20-21 0-1 0-1 0 -2 0-1 1-3 4-3 3 0 4 2 4 4l0 3 11 0 0-3c0-10-9-11-15-11zm265 11 0 24.5c1 16 29 13 29 0l0-15-13 0 0 6 4 0 0 8c-1 4-10 4-10 0 0-8 0-16 0-23 0-4 8-4 9 0l0 4 10 0 0-4.5c-1-15-29-13-29 0zm-120-11c-12 0-17 4-16 15 2 9 19 14 20 17l0 3c0 1.5-1 3-4 3-3 0-4.5-2-4.5-4l0-4-12 0c-1 12 8 16 16 16 10.5 0 16-3 16-13 1-14-18-15-20-21 0-1 0-1 0 -2 0-1 1-3 4-3 3 0 4 2 4 4l0 3 11 0 0-3c0-10-9-11-15-11zm-75 1l-1 44h11l0-40l8 40h11l7-40v40h11l-1-44h-18l-5 34l-5-34h-18zm101 0v32c-0 1 0 2 0 2 1 8 7 11 16 11 8 0 15-3 15-11 0 0 0-2 0-3v-31h-11v32c0 1 0 1 0 2 0 1-1 3-5 3-3 0-4-2-4-3 0 0 0 -1 0 -2v-32zm-139 0l-8 44h12l6-40 6 40h12 l-8-44zm182 0v44h11l0-36 11 36h16v-44h-11l0 35-10-35z"/></svg>
```
edent: I *think* we can use a transformed ellipse here.  Something like
```
<ellipse transform="matrix(1 -.2 .2 1 0 0)" cx="190" cy="300" rx="220" ry="60"/>
```
That's not quite right, but with a bit of tweaking it should fit.

edent:
I love all of these. If you can use the ellipse on the swoop - I'll merge them all.

logiclrd:
I wasn't sure if it was exactly an ellipse, but sure enough :-) _73 bytes later_

logiclrd:
(Rebased to resolve conflict in `images/reference/index.md`.)

edent:
Argh! So frustrating that the swoop is still a few bytes too large.

I hereby summon @judeGibbons and @baybal to see if they can work some magic on in :-)

logiclrd:
One other question, I suppose -- since there are 3 different versions, do you want to feature them all in `README.md`? Currently only the full logo w/o swoop is featured.

edent:
They each should be in the README as separate entries. I'm happy to accept a PR with just the two <1024 bytes. Then we can work on the swoop separately.

images/svg/samsung_swoop.svg
baybal: You can use two arcs in a single path

logiclrd: Ah, sweet :-) Minor issue, though, the SAMSUNG part (unintentionally) protrudes very slightly past the edge, and with the new approach, that protrusion is now rendered in blue. I think the ellipse needs to be made just one or two pixels larger, so that the top-left and bottom-right of the text are fully contained within it.

logiclrd: I have created PR #464 to address this.

Issue #376, opened 2020-01-16
Add Friendica


nupplaphil:
* Name: Friendica
* URL: https://friendi.ca
* Link to icon: https://github.com/friendica/docker/blob/master/friendica.svg

edent:
Hey Hacktoberfest participants! If you'd like to tackle this issue, [please follow the simple guidelines](https://github.com/edent/SuperTinyIcons/blob/master/CONTRIBUTING.md).

logiclrd:
I'll give this one a go. :-)

PR #456, opened 2020-10-20
Add Friendica logo (SVG)

logiclrd:
This PR adds `friendica.svg`, the Friendica logo at 880 bytes. :-)


logiclrd:
Resolves #376.

edent:
Love it! If you remove the newlines, you can save even more.

PR #464, opened 2020-10-21
Fix minor visual flaw in samsung_swoop.svg

logiclrd:
The Samsung with blue "swoop" ellipse icon that was recently merged has a minor visual flaw. It was made to fit into 1KB by merging the swoop with the text, but the text's exact size and position meant that it extended past the edge of the ellipse ever so slightly, and with the new approach, that part now renders in blue, as it represents an area of odd crossings. This PR tweaks the arcs for the swoop ever so slightly so that the text is definitely within the swoop, and has pretty close to identical (minimal) spacing between the edge of the letter and edge of the swoop in both the top-left and bottom-right.

The file is still 935 bytes long.


edent:
You two are GENIUSES!

PR #468, opened 2020-10-26
Update twitter.svg

magnoliaceiling:
Changed hex to match blue in twitter brand guidelines: https://about.twitter.com/content/dam/about-twitter/company/brand-resources/en_us/Twitter_Brand_Guidelines_V2_0.pdf


edent:
Good spot!

Issue #365, opened 2019-11-12
Validate SVGs


edent:
Using https://validator.w3.org/check shows some errors in SVGs.

For example https://github.com/edent/SuperTinyIcons/blob/master/images/svg/itunes_podcasts.svg

> Element stop is missing required attribute offset.

Doesn't affect rendering, as far as I can tell. But might be worth checking all of them.

romain-dartigues:
Should it be part of a meta-task (GitHub project?) about continuous integration?
Some checks would be informational (like the SVG validation against the W3C standards), other would be critical (file size, missing reference, ...).

jonatjano:
I started working on this one (didn't push it yet),
I tested for size and W3C validation, there are 8 validation failures, gonna fix these before creating a pull request (likely tonight).

I did the tests with [mocha](https://mochajs.org/) + [chai](https://www.chaijs.com/) and installed [node-fetch](https://www.npmjs.com/package/node-fetch) for easiness (all as devDependencies), tell me if you prefer another framework.

Issue #50, opened 2017-11-26
add accessibility guidance


LJWatson:
Would it be helpful to provide some short guidance on adding accessibility semantics to the icons (when used within links and/or without)? Happy to suggest something if so.

edent:
Oooh! Yes please. At the moment I'm just using `alt` - but happy to know of better ways.

LJWatson:
It'd be good to add `role="img"` and `aria-label="iconBrand"` to each of the super-tiny SVG files. Then reference this from the readme.md. Will this bust the super-tininess?

The advice on the readme would then be something like:

<The super-tiny icons are accessible by default. Each one has:
* `role="img"`, to inform screen readers that the icon is a graphic
* `aria-label="XYZ"` (where XYZ is the icon's brand name), to give the icon an accessible name

Note: if using the ```<svg>``` elements within an ```<img>``` element, the alt attribute should still be used to give the icon an accessible name because the ARIA is not recognised in this context.<

If this seems like a good approach, I'll get a PR together that adds the ARIA to the super-tiny SVG files, and the above information to the readme.

edent:
Sounds good! Looks like that will add about 40 bytes to each logo. That'll bust the size of a few of them, so we might have to do some fiddling.

PR #471, opened 2020-10-27
Tests

jonatjano:
based on #365 

test for each svg if the size is under 1KB and if the xml is accepted by the W3C validator

I had to dynamically remove the `aria-label`s because the validator return `error: Attribute “aria-label” not allowed on SVG element “svg” at this point.` (only for the test, the attribute is not removed from the actual svg)

I used [mocha](https://mochajs.org/) + [chai](https://www.chaijs.com/) and installed [node-fetch](https://www.npmjs.com/package/node-fetch) for easiness (all as devDependencies)


edent:
That's some amazing work! I'm going to have to test them in a few browsers to see how they render.

Well done! And welcome to the team.

jonatjano:
Hello,
I'm happy you like it, if they break in any browser tell me so I can do the changes,
In case this PR takes some time to be accepted could you give it the `hacktoberfest-accepted` label please.

Hope we can do some great things as a team :slightly_smiling_face:

edent:
With pleasure :-)

edent:
A few minor changes, and we should be good to go!

Apple Music doesn't render properly in Firefox. See the original discussion: https://github.com/edent/SuperTinyIcons/pull/190
Screenshot_2020-10-28 Tests by jonatjano · Pull Request #471 · edent SuperTinyIcons

You've removed `xlink` on a few of them. Sadly, I think it is needed for backwards compatibility. See this discussion: https://github.com/edent/SuperTinyIcons/issues/184

Other than that - I'm *thrilled* with these :-)

btrem:
> 
> I had to dynamically remove the `aria-label`s because the validator return `error: Attribute “aria-label” not allowed on SVG element “svg” at this point.`
> 

Is aria-label necessary? It seems to me that the `<title>` is more appropriate, and valid. When I use a super tiny svg icon, I replace the aria attribute with the element, and it seems to work just fine. BICBW.

jonatjano:
> Apple Music doesn't render properly in Firefox. See the original discussion: #190
TIL I need to test in browser because ubuntu's image viewer seems to not render gradients

> You've removed `xlink` on a few of them. Sadly, I think it is needed for backwards compatibility. See this discussion: #184
I'll add it back

edent:
@btrem the suggestion to add the `aria-label` was by accessibility expert @LJWatson. You can read the original discussion at https://github.com/edent/SuperTinyIcons/issues/50 

@jonatjano welcome to the confusing world of SVG rendering!

jonatjano:
Ok, so I took a further look, ubuntu's image viewer do show gradient, but they need all attributes to be set correctly.

e.g. `offset=""` won't work while `offset="0"` works, 

This in addition to the xlink, some icons will go over 1KB, I'll try to refactor them.

If we support these stricter viewer, it could be an idea to see what kind of test could be added.

------------------------
For the case of apple music, it feels like they changed their logo : on their [styleguide](https://www.apple.com/itunes/marketing-on-music/identity-guidelines.html#apple-music-icon) the icon looks like this :
image

edent:
Oh! That makes life easier. I'm happy if you want to update to the official Apple icon.

btrem:
> the suggestion to add the `aria-label` was by accessibility expert @LJWatson. You can read the original discussion at #50

I've read that thread. At the time, AIUI, there was no `<title>` element, nor a label. So he was suggesting aria-label to provide something. My point is that if you replaced `aria-label="foo"` with `<title>foo</title>` you would get valid markup **and** accessibility vis-à-vis labels. It would also add two bytes, so you'd want to take that into consideration.

jonatjano:
Actually, that's only one byte since `aria-label` is always preceded by a space which is not needed anymore if we use title.

But according to this [document from the w3c about svg accessibility](https://www.w3.org/TR/svg-aam-1.0/#mapping_additional_nd), both `aria-label` and `title` are as good. (There is a TL;DR; in the green box at the end of the section)

The only difference is when opening the svg as standalone, the `<title>` is recognised as the tab name on browser while `aria-label` doesn't.

There are currently 3 SVGs with a size of 1023 bytes where 1 additionnal byte is a problem but they can be optimized : 
- firefox has a line feed we can remove, 
- minecraft use some complex transforms, I'm not sure if it is possible to simplify them
- samsung has an unused `fill` attribute on the `<svg>` element (every child as it's own `fill`)

jonatjano:
maybe we should move the `aria-label` vs `<title>` discussion to another issue ?

edent:
I'm accepting these. I'll test them in a few other renderers.
Please open a new issue to discuss the title issue. 
Thanks :-)

PR #473, opened 2020-11-05
Add Element icon

zakm:


edent:
This is BRILLIANT! Thanks @zakm.  I have two suggestions.

1. Take a look at reducing the precision. Some of your paths have 4 decimal places. They can probably be dropped.
2. I wonder if having one path and the rotating it would shrink the file even more?

zakm:
@edent Re-using the one path did bring the file size down a bit, I didn't realize that was possible, thanks!

edent:
Boom! Love it!

Issue #478, opened 2020-11-08
Extra reduction in size


yatru:
You can use the (https://yqnn.github.io/svg-path-editor/) to reduce the path size even more (In some cases).
Source code: https://github.com/Yqnn/svg-path-editor/

edent:
Cool! Please let us know how you get on.

yatru:
You only need to go to https://yqnn.github.io/svg-path-editor/ and paste the Path and click on Minify
In some cases it will not help, but in other it will. The source code is available to automate this. 
Also, it's an editor, so you can personalize the logos.

edent:
OK. Which icons need the most work?

yatru:
They are pretty well, but for example twitter could be reduced 19 bytes
"M437 152a72 72 0 01-40 12a72 72 0 0032-40a72 72 0 01-45 17a72 72 0 00-122 65a200 200 0 01-145-74a72 72 0 0022 94a72 72 0 01-32-7a72 72 0 0056 69a72 72 0 01-32 1a72 72 0 0067 50a200 200 0 01-105 29a200 200 0 00309-179a200 200 0 0035-37" 

You should check the source code to automate the possible reduction in some cases

yatru:
Slack only 2 bytes "M149 305a39 39 0 01-78 0c0-22 17 -39 39 -39h39zM168 305a39 39 0 0178 0v97a39 39 0 01-78 0z"

edent:
Thanks. Please send that as a PR.

If you know of a way to automate with GitHub's CI, please let me know.

yatru:
Thanks to you , great collection

TPS:
> Thanks. Please send that as a PR.
> 
> If you know of a way to automate with GitHub's CI, please let me know.

@edent You should subscribe to [Imgbot.Net](httpS://imgbot.net/), which would automatically shrink, commit to `imgbot` branch, & PR most images (not just SVGs).  You'd also probably want to commit an `.imgbotconfig` in your root to make sure to squeeze every bite out of the already tiny SVGs.  I use
````
{
    "compressWiki": "true"
    "minKBReduced": null
}
````
If you wish, I could submit a PR that'd have that `.imgbotconfig` & the 1st round of SVG savings.  Just let me know!

edent:
@TPS we have exhausted the limits of SVGO. That's all imgbot can do for SVGs. While it is great for unoptimised files, it can't do anything for our already optimised files.
[...]

PR #482, opened 2020-11-09
ImgBot savings

TPS:
@edent As requested in https://github.com/edent/SuperTinyIcons/issues/478#issuecomment-723659800, this is what [ImgBot](www.ImgBot.net) has distilled. Each of the included files is just a few bytes smaller (mostly whitespace), so I don't know how welcome _that_ is.  Also, I've included a custom `.imgbotconfig` which specifically excludes the `reference` folder.  Let me know what you think! 🙇🏾‍♂️


TPS:
I'll go through when I can & check for what savings other than whitespace, but why insist on such?  It's much easier to check before/after if _both_ are with/without such, &, @ these sizes, removing whitespace tends to be significant.

edent:
I don't think I was clear enough. A `\n` takes up the same number of bytes as a space. So we use newlines for readability. That also makes it easier to see what a diff has done.

I've just pushed some changes which removed Windows style newlines from some files, and trailing newlines from the end of some files.

I'd be interested to see whether imgbot makes any file savings against these newer, smaller files.

Thanks for your work on this, it is appreciated.

TPS:
Eyeballing it, it looks like it got some missed leading `0.`→`.` & similar.  I'll have ImgBot regenerate against the repo's current iteration asap.

TPS:
Replaced by #485

PR #485, opened 2020-11-13
Imgbot savings, v2

TPS:
@edent This replaces #482 to account for latest minimal repo SVGs.  This still [saves](https://github.com/edent/SuperTinyIcons/issues/478#issuecomment-723659800) ~40 bytes over 15 SVGs, again mostly whitespace, but the odd leading `0.`→`.`, &c, conversion.  Thoughts?


edent:
Thanks. Because of the formatting, it'll take me a little time to go through these.

TPS:
@edent A couple thoughts:
- I'd like to open an issue upstream w/ SVGO &/or ImgBot re: the newline issue.  I understood what you meant by the replacing the `CR LF` line endings w/ `\n`.  Technically most of the time, [`CR`=`\r` & `LF`=`\n`](https://en.m.wikipedia.org/wiki/Newline), so I think there might/should be an option to preserve the newline while not wasting bytes.
- Once that's resolved, would you be interested in ImgBot?   I realize _this_ PR was onerous, but I think it's because it was so many files & the above `\n`problem.  If so, it wouldn't be bad to merge the `.imgbotconfig` _before_ subscribing.  @ worst, it creates only PRs & an `imgbot` branch, anyway — _no_ automatic merges.

edent:
Yes, very happy to have imgbot running - if it can preserve the existing newlines.

This is a brilliant resource and I want to thank you for taking the time to submit these PRs.

TPS:
I've opened https://github.com/svg/svgo/issues/1300 re: the `\n` (& I threw in `\t` for good measure) concern.

PR #486, opened 2020-11-14
ImageBot savings, v3

TPS:
@edent This is probably my last _ImgBot_ PR, as these look to be last bytes left to save that way _en masse_.  The rest can be handled by your eventual subscription. I do have a few more PRs in mind via other tools for the SVGs & Android drawable XMLs, & also the JPEGs, GIFs, & PNGs in mostly the `/reference` folder, since it seems dishonest to compare optimized vector to ***un***optimized raster images.


TPS:
@edent It occurs to me that many `CR`|`LF`|`Tab`s might be _eliminated_ outright to save bytes,  in which case the _only_ solution would be to awkwardly, randomly convert some _required_ whitespace to `LF`s just to break up long 1-liners.  Do you have rules about such?

TPS:
@edent No, I mean, much of the time, _no_ whitespace is required by the spec @ place that would make logical sense to a human parsing the file, so that's extraneous byte(s).  Is there a rule this repo has about when to use `space`, `tab`, or `LF` when whitespace is _required_ by the spec?

TPS:
Here's an example of what I mean.  I reworked the `guideline.svg` via 80 _(arbitrary)_ characters/line into:
```svg
<svg xmlns="http://www.w3.org/2000/svg" aria-label="Guidelines" viewBox="0 0
512 512"><rect width="512" height="512" rx="15%" fill="#222"/><circle
fill="#f43" r="256" cx="256" cy="256"/><circle fill="#fd0" r="217.5" cx="256"
cy="256"/><circle fill="#5d6" r="192" cx="256" cy="256"/></svg>
```
(Note `LF` inside tags & even strings.)  Is this acceptable for the repo?  _(As an aside, I ran into `<svg`…`role="img"`…`>` in some of the SVGs, which seems unnecessary.  Is it needed for anything?)_

TPS:
The PR broke due to keeping my fork in tune with the main, so closing.

PR #487, opened 2020-11-14
Shrinks ubuntu.svg

btrem:
Uses 1 circle with stroke for main ring instead of 2 separate circles.
Simplifies path.


btrem:
I missed that optimization. Clever! I updated the `fill` attribute in my branch.

PR #492, opened 2020-11-16
Adds WHATWG icon

btrem:


btrem:
I had to fix the aria-label, hence the push after the pull request. Sorry for that. It is now ready to be merged.

PR #493, opened 2020-11-17
FileOptimized Raster files

TPS:
[As mentioned](https://github.com/edent/SuperTinyIcons/pull/486#issue-521044395), here're the [FileOptimize](https://nikkhokkho.sourceforge.io/static.php?page=FileOptimizer)d raster files in the repo for a more honest comparison to the optimized SVGs, along with an adjusted `.imgbotconfig` to handle these on an ongoing basis hereafter (whenever you might subscribe to ImageBot).  This certainly makes the filesize comparison less stark, but SVGs have _many_ more benefits than just that!


edent:
Ah! The reference page is just to see how the icons **look** vs the originals. I hadn't put sizes on there - some of them aren't square. Let me have a think about what I want to do with them.

TPS:
I was thinking of the ["How Small?" section of the Contributing doc](https://github.com/TPS/SuperTinyIcons/blob/master/CONTRIBUTING.md#how-small) , since you do invite filesize comparison there.

TPS:
@edent So, you're not interested in this?

edent:
I'm happy to take smaller versions of the "how small" section. But the reference images as just straight copies from official sources.

PR #496, opened 2020-11-24
Smaller microsoft.svg

zfrohardt:
By using <path> elements instead of <rect> elements, microsoft.svg can be made 56 bytes smaller


edent:
Nice!

PR #497, opened 2020-11-24
Reduced size of thisamericanlife.svg

zfrohardt:
By replacing a collection of <rect> elements and a <polygon> element with a single <path>, the file size of thisamericanlife.svg can be reduced from 443 bytes to 309 bytes.


edent:
👍

zfrohardt:
Haha I realize now that my comment does not make much sense because I did not escape some angle brackets... Here is the original message with some escaping:

By replacing a collection of \<rect\> elements and a single \<polygon\>, the file size of thisamericanlife.svg can be reduced from 443 bytes to 309 bytes.

PR #498, opened 2020-11-25
Shrink the size of thisamericanlife.svg by optimizing path

zfrohardt:
By rearranging the order of commands in the \<path\> element, we can shrink the overall size from 309 bytes to 302 bytes.


zfrohardt:
My apologies, I am using this repository as a way to learn how to contribute to open source projects, and I did not realize that by pushing to master I would be adding commits to this pull request. I will make feature branches from now on

edent:
That's OK - we all have to learn somewhere.

Could you please check the Vivino one against the reference icon? Your new version looks different from our original.

zfrohardt:
From what I can tell, my version is closer to the official reference file than the one that I modified. In the previous version, there were lots of misalignments peppered throughout. 

If you look at the reference file, all of the circles are placed along a sort of standardized grid system, where each circle is equidistant to each of their closest neighbors. Each circle is exactly aligned with every other circle in the same row and column. My version replicates this design and fixes some of the sloppiness in the previous version

zfrohardt:
Would you like me to make a separate pull request for altering vivino.svg? The current icon can be shrunk without any changes to the icon. I just also happen to think that the current icon's fidelity can be improved

edent:
Sounds good.

PR #499, opened 2020-12-02
Update elementaryos.svg (767 Bytes => 477 bytes)

kpym:
Use circle and simplified path.


edent:
Brilliant stuff! You can shrink a few more bytes from it by following the guide in the CONTRIBUTING files.

PR #501, opened 2020-12-03
Fixed an error in the size description of the icon

taskinoz:
The Table that show the size difference between the Flickr icons are different in the Size Comparison table and the Social Media table.

I just changed the the size from 251 to 250 as is says in the social media table and what the Windows file system reports.


edent:
Good spot! Thanks :-)

PR #507, opened 2020-12-05
brave browser icon

rrichy:
this should be on browser category


edent:
Looks great. Could you follow the template in https://github.com/edent/SuperTinyIcons/blob/master/CONTRIBUTING.md

PR #514, opened 2021-01-04
Clarify problem with rounded corners in Android Studio

ziegenberg:
Android Studio works fine with rounded corners; it just doesn't like percentage length values (or any other values with a unit identifier). Any of the following optional unit identifiers would prevent converting SVGs to Android Vector Drawables: `("em" | "ex" | "px" | "in" | "cm" | "mm" | "pt" | "pc" | "%")`.

I used 77 as a close approximation for 15% of 512.


edent:
Thanks. Wonder if we should just change them all to 77?

ziegenberg:
Changing all rx values to 77 is up to you. I doubt that Google will fix the percentage length values for svg2vector conversion soon.

PR #523, opened 2021-02-18
Add Overleaf

Eiim:
First submission, hope I got everything right!

Adds icon for [Overleaf](https://www.overleaf.com) in 452 bytes.


edent:
Brilliant - thanks!

Issue #503, opened 2020-12-03
Malt icon


neilime:
## The icon I want is:

* Name: Malt
* URL: https://www.malt.fr/
* Link to icon: https://res.cloudinary.com/hopwork/image/upload/v1599481800/rebranding2020/malt-logo/icon-180x180.png

Thank you

edent:
Thanks! Do you know if there's an SVG version of the logo?

neilime:
I didn't found it

neilime:
I found it:  https://res.cloudinary.com/hopwork/image/upload/rebranding2020/malt-logo/malt-red.svg

PR #526, opened 2021-04-02
Add Malt logo

zfrohardt:
Solves issue #503 by adding malt.svg into the repository. There is still some optimization that can be done to minimize the file size, but this is an acceptable start.


edent:
Oooh! Just squeezed it in to the filesize :-)

PR #531, opened 2021-05-10
Update Android icon to new branding

AnalogCyan:
Google updated the Android brand in 2019; SuperTinyIcons still uses the old branding.

* [Brand guidelines](https://developer.android.com/distribute/marketing-tools/brand-guidelines.html)
* [Brand update blog post](https://blog.google/products/android/evolving-android-brand/)
* Icon size: 741 bytes

While the updated icon is within the size requirements, it's bigger than the old icon and I'm fairly certain can be optimized further (I might take a crack at doing that at a future date if nobody beats me to it).

First submission here, hope I did everything correctly!


edent:
Thanks! Looks brilliant. Please can you update the README with the correct filesize and I'll merge.

PR #532, opened 2021-05-10
Update tiktok.svg

kpym:
This file is bigger by 38 bytes (446 => 484) but it avoids the use of `<!ENTITY ...>` and `<!DOCTYPE ...>` and as consequence:
- It respect the template starting with `<svg xmlns="http://www.w3.org/2000/svg" ...>`.
- This `svg` code can be integrated directly in `html`.


kpym:
`tiktok.svg` is the only icon that has `<!DOCTYPE>` before the `<svg>` tag.

edent:
Brilliant stuff. Please update the README and I'll merge.

Issue #509, opened 2020-12-09
Add "Text editors" section


kpym:
It would be nice to add some icons of the most popular text editors.

Here is my proposal for Sublime Text's icon (728 bytes):

```svg
<svg xmlns="http://www.w3.org/2000/svg"
aria-label="Sublime Text" role="img"
viewBox="0 0 512 512"><rect
width="512" height="512"
rx="15%"
fill="#424242"/><rect transform="skewY(-17.5)" x="110" y="375" width="292" height="107" rx="8" fill="url(#lg)"/><rect transform="skewY(17.5)" x="110" y="122" width="292" height="107" rx="8" fill="#ff9800"/><rect transform="skewY(-17.5)" x="110" y="191" width="292" height="107" rx="8" fill="#ff9800"/><defs><linearGradient id="lg" x1="174" x2="189" y1="725" y2="716" gradientTransform="scale(1.65 .605)" gradientUnits="userSpaceOnUse"><stop stop-color="#ff9700" offset="0"/><stop stop-color="#f48e00" offset=".5271813"/><stop stop-color="#d06f00" offset="1"/></linearGradient></defs></svg>
```
![sublime_text2](https://user-images.githubusercontent.com/3142701/101671858-610ff900-3a55-11eb-8601-efd878f95862.png)

edent:
That looks amazing! Very happy to receive a pull request with this.

kpym:
Solved pull request #534.

PR #534, opened 2021-05-10
Create sublimetext.svg

kpym:
Add sublimetext.svg icon.


kpym:
This is a fix for #509.

Issue #535, opened 2021-05-12
The S in SVG is duplicated


benkasminbullock:
In svg.svg, the S shape is duplicated.

In the following path,

M191 318a35 35 0 10-35 35 35 35 0 11-35 35m171-125l-36 172-36-172m-29 55a35 35 0 10-35 35 35 35 0 11-35 35m270-70a35 35 0 00-35-35 35 35 0 00-35 35v70a35 35 0 0035 35 35 35 0 0035-35v-35h-35"/><path d="M64 256h384v12H64"

the section from after M191 318 to the following m171-125 is a duplicate and can be omitted completely.

PR #536, opened 2021-05-12
Remove duplicated S from svg.svg

benkasminbullock:
https://github.com/edent/SuperTinyIcons/issues/535


edent:
Good spot! Please update the README with the new size, and I'll merge.

PR #541, opened 2021-05-13
Fix broken image links on the references page.

AnalogCyan:
Many of the icons on the [references page](https://github.com/edent/SuperTinyIcons/blob/master/images/reference/index.md) had the wrong file names resulting in the icon not being shown.

This fixes all but the Verizon icon, since that one doesn't appear to have any icons here at all.


edent:
Thanks!

PR #543, opened 2021-05-19
Remove xlink:href

benkasminbullock:


edent:
OK - this looks good. But I think you can go one step further and remove `xmlns:xlink="http://www.w3.org/1999/xlink"` from each file as well. That'll save even more space.

PaulCapron:
**Caution:** it’s **only in SVG 2** that a ‘naked’ `href` attribute can be used, instead of `xlink:href`.

Most browsers do support SVG 2, but many other programs do not (yet).

For instance, here is how the modified (using a naked `href`) Amazon S3 logo is seen, _badly_, by Ubuntu/Gnome image viewer:
Capture d’écran de 2021-10-07 13-01-33

PaulCapron:
Note that the `http://www.w3.org/1999/xlink` namespace can be ‘nicknamed’ as anything, it isn’t required to be `xlink`.

I.e.:

    <svg … xmlns:x="http://www.w3.org/1999/xlink">
      <use x:href="…"/>

is perfectly valid. Here `http://www.w3.org/1999/xlink` is ‘nicknamed’/referenced as simply **`x`**.

And that would save a couple bytes, without impacting compatibility with programs understanding only SVG 1.

PR #551, opened 2021-06-16
Update css3.svg

allredmatt:
Typo in aria label changed from HTML5 to CSS3


edent:
Looks good - can you also please update the file size in the README?

Issue #555, opened 2021-06-27
Briar App


edent:
## The icon I want is:

* Name: Briar
* URL: https://briarproject.org/
* Link to icon: https://briarproject.org/img/briar_logo_large.png

* Vector version of the logo: https://code.briarproject.org/briar/briar/-/blob/master/briar-android/artwork/logo_circle.svg
* Official style guide or brand guide:

PR #558, opened 2021-07-17
added Briar svg and reference

zfrohardt:
Adds Briar Project (https://briarproject.org/) to the repo and solves issue #555


edent:
Brilliant! Thanks

PR #563, opened 2021-07-18
Minimized Jacobin to 241 bytes

zfrohardt:
The path in jacobin.svg was unoptimized, so I shaved off a few bytes


edent:
I'm granting you commit access to the Repo.  This is per the advice of [The Pull Request Hack](https://felixge.de/2013/03/11/the-pull-request-hack.html).

Please use this privilege wisely.

PR #568, opened 2021-07-19
Shrank Npm to 299 bytes

zfrohardt:
I shaved 2 bytes off of npm.svg by rearranging its path commands


edent:
This is what we like to see!

PR #581, opened 2021-08-10
Added Proxmox (701 bytes)

Emersont1:
Hi, I've added the proxmox logo (https://www.proxmox.com/en/), which comes in at 701 bytes

Here's a copy of the logo and a second with the guides for reference

proxmox logoproxmox logo with guide

I currently don't have access to android studio so am unable to make the android versions


edent:
Hiya. Looks good. Would you mind formatting it in the same way as the others? Thanks!

PR #582, opened 2021-08-10
Added Apache Guacamole (1013 bytes)

Emersont1:
Hi, I've added the logo for apache guacamole (https://guacamole.apache.org/), which comes in at 1013 bytes
I currently don't have access to android studio so am unable to make the android versions

guacamole logo


edent:
Hiya, sorry for the delay. Looks really good - but can you format it like the template in https://github.com/edent/SuperTinyIcons/blob/master/CONTRIBUTING.md#new-file ?

PR #584, opened 2021-09-18
Update Laravel

### LINKED ISSUE #562: Laravel icon is out of date
It appears that [Laravel](https://laravel.com/) has updated its branding, and the logo in this repo is now out of date. I briefly looked at the new logo, and it shouldn't be too hard to get it down to size, but it will be larger than the existing one we have

Eiim:
Fixes #562 

Actually saved a few bytes off the previous version!


edent:
If you remove the newlines - like the template - you'll save even more.

Eiim:
That I will!

PR #588, opened 2021-10-01
Add PornHub logo with a size of 850 bytes

GearWalker:
I have revised all tags before submit.


edent:
Looks really good - but can you format it like the template in https://github.com/edent/SuperTinyIcons/blob/master/CONTRIBUTING.md#new-file

PR #589, opened 2021-10-02
Added Manjaro at 375 bytes

montylion:
- Added Manjaro icon at 375 bytes
- Edited readme
- Added reference image
- Edited reference page
- Added Android image converted with [svg2android](https://inloop.github.io/svg2android/)


edent:
Thanks!

Issue #605, opened 2021-10-09
Add Dart Icon


ecrax:
## The icon I want is:
* Name: Dart
* URL: https://dart.dev/
* Link to icon: https://icons8.de/icon/7AFcZ2zirX6Y/dart

* Vector version of the logo: https://commons.wikimedia.org/wiki/File:Dart_programming_language_logo.svg
* Official style guide or brand guide: https://dart.dev/brand

PR #606, opened 2021-10-09
Added Dart Icon (691 Bytes)

ecrax:
- [x] Add Origin svg at 691 bytes - closes Add Dart Icon #605  
- [x] Updated Readme
- [x] Add reference image
- [x] Edit reference page
- [x] Add Android Image


edent:
Good stuff. I reckon you can reduce the precision to save some bytes.

PR #618, opened 2021-10-20
Tweak git ignore and attributes

ace-dent:
Smoother commits:
- Add system files to `.gitignore`.
- Create `.gitattributes` to force line endings.


ace-dent:
Would appreciate a review of `.gitattributes`, in case there's any unforeseen issues with explicitly declaring files as `text`.

ace-dent:
Made a little mess... see #624.

PR #624, opened 2021-10-30
Fix line endings

ace-dent:
15 files need `CRLF` line endings changed to `LF`.


ace-dent:
This fixes issues highlighted by #618... that may leave other's scratching their heads :-/

PR #636, opened 2022-03-30
Optimize Samsung logos

Eiim:
Optimizes all three Samsung logos, especially the main one, which was previously at a high 1023 bytes.


edent:
Amazing, thanks!

PR #639, opened 2022-04-10
a few bytes less for Chrome Icon

3l3gant-cod3s:
1) replace yellow path with circle
2) reuse Chromium icon’s shorter path


PaulCapron:
Good job! Thanks, merged ;-)

PR #640, opened 2022-04-11
Add nixos logo (544 bytes)

lunik1:
- [x]  New file which is under 1,024 bytes (544 bytes)
- [x] Edit Readme
- [x] Add reference image
- [x] Edit reference page
- [x] Android image

### DIFF
--- a/images/svg/nixos.svg
+++ b/images/svg/nixos.svg
@@ -1,0 +1,9 @@
+<svg xmlns="http://www.w3.org/2000/svg"
+aria-label="NixOS" role="img"
+viewBox="0 0 512 512"><rect
+width="512" height="512"
+rx="15%"
+fill="#fff"/>
+<g fill-rule="evenodd">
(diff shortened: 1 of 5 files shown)


images/svg/nixos.svg
```diff
@@ -0,0 +1,9 @@
+<svg xmlns="http://www.w3.org/2000/svg"
+aria-label="..." role="img"
```
PaulCapron: Ideally the `aria-label` should not be a placeholder ;-P But something like, `aria-label="NixOS"`. 

This is for accessibility, see e.g. https://developer.mozilla.org/en-US/docs/web/accessibility/aria/attributes/aria-label

PaulCapron:
Hi @lunik1,

Made a couple minor comments, if you could look into them that would be top. Great job already!

PaulCapron:
Thanks, that will do! Merged :-)

PR #645, opened 2022-04-15
Update average size

Eiim:
It's been a while since that average size number has been updated, so I figured it would be good to recalculate it.


Eiim:
Since there have been a bunch of optimizations by me and @3l3gant-cod3s in the past week, if those changes are merged first I'll go ahead and update this accordingly , as that will likely shift the average by at least few bytes.

PR #660, opened 2022-07-18
Update guacamole.svg (1021B → 697B)

kpym:
Not only is the size smaller but the icon is more accurate.


PaulCapron:
Hi,

Great job!
Could you please also update `README.md` to reflect the new size?
The new size by the way seems to be 698 bytes, not 697. A newline got added to the end of the file; ideally it should be removed!

PR #662, opened 2022-08-01
Added Floatplane

PersonMeetup:
Adds the [Floatplane](https://floatplane.com/) logo with everything but an Android version included.

The only problem I see with this PR is the size of the SVG, which comes at 1019 bytes. I was having a hard time reducing the file size while also maintaining image accuracy, but I think I managed it.


Eiim:
Looks good to me, but if you want to make it a bit smaller fairly easily you can merge all the paths into one and put the fill=#fff on that rather than having them be in a group.

PersonMeetup:
Thanks for the tip @Eiim; that dropped it down to 944 bytes!

PR #667, opened 2022-10-11
Fix broken entries at reference/index.md

ammendonca:


edent:
Thanks! Much appreciated :-)

PR #685, opened 2022-12-03
Updated Chrome & Chromium icons

kuubeu:
Logos updated to use the 2022 design. The colors are set according to official flat-colored versions from Wikimedia Commons


edent:
Nice one! Thanks.

PR #688, opened 2022-12-06
Update TikTok logo & correct recent changes

kuubeu:
- TikTok now displays correctly in Inkscape, in addition to being smaller (433 → 407)
- corrected file sizes: Chrome, Chromium, Atom – I think editing through GitHub's editor saves files using CRLF, which I didn't notice at the time 😞


Eiim:
Looks good!

Issue #575, opened 2021-08-02
Challenge! OpenStreetMap icon


edent:
## The icon I want is:

* Name: OpenStreetMap
* URL: https://www.openstreetmap.org/
* Link to icon: https://wiki.openstreetmap.org/w/images/7/79/Public-images-osm_logo.svg

There's a set of simplified logos at https://wiki.openstreetmap.org/wiki/Logos#Other_logos which might be easier to shrink.

For example https://wiki.openstreetmap.org/w/images/5/57/Logo_simple_colors.svg

Eiim:
I have a 1021 byte icon based off the OSM Hispanic logo, but I'd like it to look a little nicer:

![OSM_Hispanic_modified4](https://user-images.githubusercontent.com/24627058/133871315-427623ed-6d37-4c86-a7c9-5c10d6108b09.png)

[OSM_Hispanic_modified4.zip](https://github.com/edent/SuperTinyIcons/files/7189136/OSM_Hispanic_modified4.zip)

sidverma32:
@edent  Can i take this task?

edent:
Do it! 😀

PR #698, opened 2022-12-21
Add OpenStreetMap

colincameron:
As requested in #575

I've used the "OSM logo simple, for embroidery" logo from https://wiki.openstreetmap.org/wiki/Logos#Other_logos

The SVG was completely hand-crafted using the original as reference. It's pretty much the same, but I couldn't keep the text under the microscope. Maybe someone else can use the remaining 40 bytes to do so?


Eiim:
Nicely done! This has been on my wishlist for a while as an OSM contributor. You can remove the newlines after the starting `<svg>` to save a few more bytes.

edent:
@colincameron happy to merge if you can remove those newlines.

edent:
Fixed in https://github.com/edent/SuperTinyIcons/commit/e40f7cae846bf5dfec519265c4e8efeabb694d2b

Issue #696, opened 2022-12-17
New Icon: Dingoo handheld and OpenDingux distribution


Tabaluga:
Hello @edent ,

today I simplified and created two icons based on the guidelines, which I would like to submit for your collection:

## Dingoo handheld

The Dingoo, released in 2009, is a Chinese manufactured handheld and portable media player (PMP) by Shenzhen Dingoo Digital.

Out of the box, the handheld plays numerous emulators up to the fifth generation of [video game consoles][1]. Also included are the standard set of PMP features. [^1]

```svg
<svg xmlns="http://www.w3.org/2000/svg" aria-label="Dingoo" role="img" viewBox="0 0 512 512"><rect width="512" height="512" rx="15%" fill="#000"/><path fill="#fff" d="M205 81c-16-1-35 4-50 13-24 14-37 39-35 55 2 10 11 24 24 33 10 6 17 11 28 11 28-1 50-30 55-49 6-18 10-48-1-58-5-3-12-5-21-5zm116 9c-16 0-37 4-50 11-24 15-32 36-30 52 1 9 8 20 21 30 8 8 16 16 27 15 24-4 46-36 51-51 5-14 14-40 7-50-3-4-13-7-26-7zm83 69c-15 0-38 7-53 17-22 15-36 30-34 49 0 10 8 21 19 32 9 8 22 12 31 11 27-3 45-31 51-47 5-16 9-46 1-57-3-3-8-5-15-5zm-250 48c-9 0-18 1-27 5-24 13-43 33-42 65a195 195 0 0 0 110 144c24 10 64 21 94-9 30-29 23-52 14-77-9-24-1-38 4-61 5-35-24-54-50-58-26-5-48 7-59 3-8-3-23-12-44-12zm247 69c-14 0-31 4-45 11-22 12-38 35-33 53 3 10 8 19 20 30 9 7 19 10 29 10 29 1 47-26 51-45 3-18 7-44-1-54-4-4-12-5-21-5z"/></svg>
```

## OpenDingux distribution

OpenDingux (Dingoo Linux) is a modern Linux distribution for handhelds (Dingoo series, GCW-Zero, PocketGo2 series, Anbernic series ...).

It uses standard Linux interfaces as much as possible, to make it easier to port software from / to OpenDingux. [^2]

```svg
[...]

PR #702, opened 2023-01-15
Add Dingux, Everdrive, Open Dingux, Retroarch Icons

Tabaluga:
Hello @edent  ,

I edited and simplified four icons based on the guidelines, which I would like to submit for your collection:

- Dingux (821 Bytes) (#696)
- Everdrive (425 Bytes)
- Open Dingux (386 Bytes) (#696)
- Retroarch (381 Bytes)

best regards


Eiim:
Sorry this took so long to get around to. But I'm a little confused as to where these icons are coming from.

* Dingux: I see an icon like it on the Dingoonity wiki, but nothing to indicate that it's the Dingux logo. From some searches, it looks like it might actually be the Dingoo logo?
* Everdrive: I can't figure out what this is. Maybe a series of [flash carts](https://krikzz.com/)? Nothing that looks like the logo you have here that I can find though.
* OpenDingux: Appears to be written as "OpenDingux" whereas you have "Open Dingux" in the README and open_dingux in the file name. The logo seems to check out though.
* RetroArch: Name should capitalize "A" in README. No other concerns.

For Dingux/OpenDingux/Everdrive, they're so niche that I'm not totally certain if we'd want them. We don't necessarily need the logo of every random project that someone spun up in a weekend here, it seems like there should be a line somewhere. But that's not something we've ever invoked so it'd be completely up to @edent if we wanted to make such a rule. I'm not concerned about that for RetroArch, or the Dingoo representing Dingoo itself (rather than the very old and more obscure Dingux project)

Also, please add a break to the README file so that it stays at a max of 6 items per line.

Eiim:
Oh, one other thing. In general, we prefer dark on light logos, would you mind swapping the colors here? Unless any of these logos are very consistently shown as light on dark. It seems like RetroArch at least is normally dark on light though.

edent:
Yeah, they need to be somewhat recognisable / popular.

If you can show that the icons would be useful then please send one PR per icon.

PR #703, opened 2023-02-23
Correcting logos & reducing size

gucr:
Hi @edent 

Here is a PR to fix some logos where the shape was slightly off, and some optimizations to reduce file weight


edent:
These all look good. Want me to merge them?

gucr:
Sure! I deleted it because I forgot to update the logo sizes in the README. I planned to amend all the commits with the new sizes but I won't have time to do it in the next days.
If that's ok for you as is feel free to merge it :)

PR #705, opened 2023-03-02
Updating Mastodon logo

### LINKED ISSUE #693: Mastodon Icon thoughts
I was playing with the mastodon icon elsewhere and tried to change the colour/fill. It caused problems with the two paths so I investigated whether you could cut out the 'm' rather than draw it on... turns out you can! This version comes out 13 bytes larger than the current icon, so not sure if you want a PR for it or not...
mastodon2

gucr:
Updating Mastodon logo to reflect the 2022 brand update.
Optimized the shapes & added the new gradient

Closes #693


edent:
Much obliged!

PR #708, opened 2023-03-08
Various optimizations

gucr:
Optimizing & correcting a few logos

Corrections:
- Git: redrawing to be closer to reference
- Wordpress: Redrawing to be identical to official logo
- Discord: Correcting shape

Reducing the size of 15 logos by using arcs, quadratic paths & shorthands

I created a single comit for each modification to ease the eventual rollbacks but I can also squash everyting if that's better

Cheers!


Eiim:
It looks like you forgot to replace the RasPi README line. Otherwise, looks great!

gucr:
Thanks for the catch @Eiim ! I just amended the commit with the correction.

PR #715, opened 2023-03-20
Add example to Usage section of README showing CSS for sizing and shaping an icon

DavidAnson:
As requested at <https://mastodon.social/@Edent/110051245261401486>.


edent:
Love it - thanks!

PR #734, opened 2023-06-23
Minor optimizations

kuubeu:


kuubeu:
I guess I'll add the `offset` after all

Eiim:
That's interesting, no offset is valid in SVG 2 but not SVG 1.0 or 1.1. I don't think it would be worth it to add a version declaration, and I don't believe we have any other SVG 2 files at the moment. Interesting to consider in the future though.

PR #754, opened 2023-08-03
Use official Bandcamp colors

recursiveribbons:
The Bandcamp website logo actually uses an SVG so I copied the color from there


kuubeu:
I'm not sure how this passed the tests, since the README size doesn't match...

PR #758, opened 2023-09-12
Convert background `rects` to `paths`

kuubeu:
Saves 6 bytes per icon, as mentioned in #730


Eiim:
By the way it looks like you used "m0 0H512v512H0" in guideline.svg but "m0 0H512V512H0" in the logos. They're equivalent but it might make sense for them to match.

PR #761, opened 2023-10-03
Add twitter x new logo

0APOCALYPSE0:

PR #764, opened 2023-10-04
added Twitter X new logo

aulth:


jmb:
Hi, this is competing with PR #761 

Your submission doesn't quite follow the guidelines with the following layout and elements required:

```
aria-label="..." role="img"
viewBox="0 0 512 512"><rect
width="512" height="512"
fill="#fff"/>...</svg>
```

You should also be able to resolve/apply the translation and make it smaller!

PR #767, opened 2023-10-07
PeerTube icon

jmb:
Thought I'd add another Fediverse Icon! :smile:


Eiim:
Looks good! Mind if we shrink the path data a little bit first though? Try this for the triangles:

`<path fill="#211f20" d="M140.8,102.4V256Z"/><path fill="#737373" d="M140.8,256V409.6Z"/><path fill="#f1680d" d="M256,179.2V332.8Z"/>`

jmb:
I've removed spaces from the path definitions, but your suggestion won't work I think as it just draws a single vertical line for each triangle and then closes the path - it doesn't put the right most point in.

> `<path fill="#211f20" d="M140.8,102.4V256Z"/><path fill="#737373" d="M140.8,256V409.6Z"/><path fill="#f1680d" d="M256,179.2V332.8Z"/>`

PR #770, opened 2023-10-11
Script update for references and checks

jmb:
This PR updates the update_readme.py to keep a REFERENCES.md and a CHECK.html file updated.
A working example of the check html file is here: https://jmb.github.io/SuperTinyIcons/CHECK.html - it shows the SVG icon with border=50% so that any that don't fit the circle are easily recognised.

It currently looks for reference images with the same base name as the svg, then checks the images/reference/index.md for the source location and any other reference images that might have a different name. Eventually this check should probably be moved to look at the REFERENCES.md file and the index.md removed.

PR presented for evaluation with the understanding that tweaks will likely need to be made! Thanks!


edent:
I really like this idea. Looks good. What more work does it need?

jmb:
Just wanted to confirm the location of the files and that the html was good. Also, should I remove the reference/index.md and just use the REFERENCES.md file perhaps?

edent:
Sorry for the delay. I'm happy with this. If you could please rebase it, I'll merge.

jmb:
I *think* I've fixed it... but may have screwed things up completely! If this doesn't work then I can delete the PR and start a new branch copying the script over.

edent:
Yeah... let's start on a new branch 😂

jmb:
I think that's for the best! I'll play around with it when I have some time!

jmb:
Re-opened as #799 which is hopefully less messy!

PR #775, opened 2023-10-25
Add justgiving icon

drkane:
From the new file checklist

- [x] New file which is under 1,024 bytes
- [x] Edit Readme
- [x] Add reference image
- [x] Edit reference page
- ~Android Image~ - not done


drkane:
Thanks @edent!

PR #784, opened 2024-01-06
Add Apple Podcasts icon

btrem:
This pr adds Apple Podcasts logo. SuperTinyIcons already has iTunes podcasts. However, Apple has renamed iTunes to Apple Music, and renamed iTunes Podcasts to Apple Podcasts. A file name change would normally suffice, but Apple also changed the background color gradient.

This PR adds a new icon with the updated color, and leaves iTunes Podcasts for anyone who wants the old logo.


btrem:
The failed test result says that the [Apple Podcasts reference image](https://github.com/btrem/SuperTinyIcons/blob/add-apple-podcasts/images/reference/apple_podcasts.svg) is invalid:

>   1) apple_podcasts.svg
>      should be validated by the w3c validator:
>     AssertionError: expected ':5.54-5.81: error: SVG element “stop”…' to be empty
>     at /home/runner/work/SuperTinyIcons/SuperTinyIcons/test/test.js:34:38

I think whatever validator test.js is running is wrong. In xml, `<foo></foo>` is equivalent to `<foo/>`. I chedked the reference image using direct method on the [w3c online validator](https://validator.w3.org/), and it reported that the "document was successfully checked as XML!"

PR #788, opened 2024-01-28
Added Tapas

PersonMeetup:
This PR adds an icon from [Tapas](https://tapas.io), weighed at 700 bytes! SVG has already been shrunk down into one path and has been rounded to, at most, two decimal points.


edent:
Sorry for the delay. Looking at their website, they seem to use a word-logo, or this one https://tapas.io/favicon.ico

PR #804, opened 2024-09-18
Remove xlink (duplicates #543)

romain-dartigues:
Got to much time on my hands and I noticed that xlink has been deprecated, saving a few bytes here and there.

(Thanks to that, Glitch and Minecrat can go under 1kB!)

> SVG 2 removed the need for the xlink namespace, so instead of xlink:href you should use [href](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/href).
> — https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/xlink:href

This is a duplicate of #543; since then it seems it's support is more widely available.

Thoughs?


edent:
These all work in Firefox, Chrome, and Webkit. Also work in Gnome Image Viewer, GIMP, and InkScape.

Please can you run `update_readme.py` and then I'll merge.

PR #806, opened 2024-09-18
Shrink godot

romain-dartigues:
Thoughts welcome.


edent:
The Wikipedia one looks too changed to me. Godot looks OK though.

PR #808, opened 2024-09-26
Specify UTF-8 encoding in Python script

Eiim:
Python seems to assume CP-1252 or something similar on Windows, which breaks on emoji. This ensures that UTF-8 will be used.

### DIFF
--- a/update_readme.py
+++ b/update_readme.py
@@ -135,7 +135,7 @@ with open('REFERENCE.md','r+', encoding="utf-8") as f:
 print(f"REFERENCE.md updated.")
 
 #	Replace the table in the CHECK document
-with open('CHECK.md','r+', encoding="utf-8") as f: 
+with open('CHECK.html','r+', encoding="utf-8") as f: 
     file = f.read() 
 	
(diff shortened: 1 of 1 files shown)


update_readme.py
```diff
@@ -122,7 +122,7 @@
 print(f"README.md updated with {len(svg_list)} icons.")
 
 #	Replace the tables in the REFERENCE document
-with open('REFERENCE.md','r+') as f: 
+with open('README.md','r+', encoding="utf-8") as f:
```
edent: This should be REFERENCE.

update_readme.py
```diff
@@ -135,7 +135,7 @@
 print(f"REFERENCE.md updated.")
 
 #	Replace the table in the CHECK document
-with open('CHECK.html','r+') as f: 
+with open('README.md','r+', encoding="utf-8") as f:
```
edent: This should be CHECK

Eiim: Weird, wonder how those happened. Thanks for the catch.

PR #810, opened 2024-10-01
Added all logo with url 

justynigam:
I found all possible logo from official and some from wikipedia and other icon history website .
if any issue i will solve it


edent:
Heya - a good start.

Please can you:

* Remove the `.idea/` folder
* Check the Samsung S and Swoop logos. I'm not sure if they're correct.
* Teamspeak - should just be the icon
* Ubisoft should also just be the icon

Thanks!

Issue #813, opened 2024-10-01
Vegetarian Logo Update


edent:
The current vegetarian logo is outdated. It needs to be replaced with this - https://commons.wikimedia.org/wiki/Category:V-Label#/media/File:Vegetarian-mark.svg - in a green circle.

maxwithbug:
@edent trying this

Eiim:
Note: we currently have three open pull requests addressing this issue (#814, #816, #819). Currently, none of them meet the requirements in [CONTRIBUTING.md](https://github.com/edent/SuperTinyIcons/blob/master/CONTRIBUTING.md). If you're attempting to update this logo (or any other), make sure you understand the requirements, check other existing logos as guidelines, and ask if you have any questions.

maxwithbug:
can you please provide me image link with circle ?  is this https://commons.wikimedia.org/wiki/Category:V-Label#/media/File:Vegan_friendly_icon.png ok?

PR #816, opened 2024-10-01
 Vegetarian Logo Update #813

maxwithbug:
solve Vegetarian Logo Update #813


Eiim:
This is pretty close! Pay attention to [CONTRIBUTING.md](https://github.com/edent/SuperTinyIcons/blob/master/CONTRIBUTING.md) - in particular, the start of the SVG was changed so it no longer matches the required format, which also affects the size of the image. The background should also be white rather than transparent. It also comes in just over 1,024 bytes, so you might need to make some small optimizations. Thanks!

maxwithbug:
@Eiim updated that, can you review it, please

Eiim:
It's still not quite there. I'm looking for these lines:

```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```

Also, I think you accidentally edited the Vagrant logo, please remove those changes.

edent:
Fixed in https://github.com/edent/SuperTinyIcons/commit/4a370ee39d442391dcde287a1103c7a7aafecb5f

Issue #811, opened 2024-10-01
Add GeeksforGeeks Logo


ohmschrodinger:
## The icon I want is:

* Name: GeeksforGeeks
* URL: https://www.geeksforgeeks.org/
* Link to icon: https://iconduck.com/icons/80486/geeksforgeeks

* Vector version of the logo: https://github.com/ohmschrodinger/Hacktober-Fest-2024/blob/main/geeks%20for%20geeks%20linkscape.svg

ohmschrodinger:
can i work on this?

edent:
@ohmschrodinger you can start work on this. Please follow https://github.com/edent/SuperTinyIcons/blob/master/CONTRIBUTING.md

Munazir151:
can i work on this?

anu098jaiswal:
can u assign me this?

edent:
@anu098jaiswal please confirm you've read and understood https://github.com/edent/SuperTinyIcons/blob/master/CONTRIBUTING.md

ohmschrodinger:
Can someone help me out?
I've managed to reduce the file-size to 1.6kb
im not able to reduce it any further

here is the link to the SVG file
https://github.com/ohmschrodinger/Hacktober-Fest-2024/blob/main/compressed.svg

edent:
I think there are two main problems here.

Firstly, it doesn't look like the original logo.  See https://media.geeksforgeeks.org/gfg-gg-logo.svg

Secondly, the image is mirrored. You can use a handy trick called [Transform Translate](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/transform) to copy the left half to the right half.

ohmschrodinger:
Okay, i'll try it out

ohmschrodinger:
sorry but im not able to figure it out 😭
i tried using the transform thing on figma but file size increased drastically

PR #817, opened 2024-10-01
Add GeeksforGeeks Logo #811

Himanshuch8055:
The icon details are as follows:
Name: GeeksforGeeks
URL: https://www.geeksforgeeks.org/
Icon link: https://iconduck.com/icons/80486/geeksforgeeks
Vector logo: https://github.com/ohmschrodinger/Hacktober-Fest-2024/blob/main/geeks%20for%20geeks%20linkscape.svg

Issue no. #811


Eiim:
The submitted SVG doesn't currently follow the requirements in [CONTRIBUTING.md](https://github.com/edent/SuperTinyIcons/blob/master/CONTRIBUTING.md). Also, there shouldn't be any need to edit the Python script, if there is an issue with it, submit that as a separate PR.

Himanshuch8055:
Can you update and provide more detailed instructions for the CONTRIBUTING guide? It's currently difficult to understand.

Eiim:
I'm not sure what you're looking for. The problems with your SVG are:

1. It does not follow this format:
```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```

You might look at other SVGs in the repository as examples. The `fill="#fff"` can be changed depending on the logo (although here I don't think it should be), the rest should remain the same except the part you put in the ` ... `.

2. The file is 2,648 bytes, which is more than the 1,024 byte limit.

The other issue is the updates to the `update_readme.py` file, which should not generally be changed in a PR adding a new logo. If there are any issues with that script, they should be addressed separately.

Issue #793, opened 2024-05-08
Add mdBook/GitBook/MkDocs icons


Fexiven:
## The icon I want is:

* Name: mdBook, GitBook, MkDocs
* URL: https://github.com/rust-lang/mdBook
* URL: https://www.gitbook.com/
* URL: https://www.mkdocs.org/
* Link to icon: mdBook [SVG](https://github.com/rust-lang/mdBook/blob/master/src/theme/favicon.svg) & [PNG](https://github.com/rust-lang/mdBook/blob/master/src/theme/favicon.png)
* Link to icon: GitBook [Assets](https://gitbook.gitbook.io/brand-assets)
* Link to icon: MkDocs [PNG](https://gitlab.com/to-be-continuous/mkdocs/-/blob/master/logo.png)

* Vector version of the logo: **see above**
* Official style guide or brand guide: **see above**

Vidhi154:
sir, can i take this issue ?

Samarth1719:
Can you assign this issue to me?

edent:
@Samarth1719 did you read https://github.com/edent/SuperTinyIcons?tab=readme-ov-file#hacktoberfest

Samarth1719:
@edent yes sir.

edent:
@Samarth1719 then why are you asking to be assigned?

Samarth1719:
@edent This issue was closed?

edent:
@Samarth1719 what's does it say at https://github.com/edent/SuperTinyIcons/blob/master/CONTRIBUTING.md#hacktoberfest---important

You need to actually read the README and the Contributing Guidelines.

Samarth1719:
@edent I am beginner in this contribution so just I am trying to understand the environment

edent:
@Samarth1719 It is very simple. You need to *read*.

The README file contains concise instructions. It tells you to read the CONTRIBUTING document. 

You need to *read* both of these. That's the same for every project on GitHub. Read first, then ask questions if you don't understand something.

PR #827, opened 2024-10-02
Added mdBook/GitBook/MkDocs icons #793

HemanthTigga:

```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```


HemanthTigga:
please review

edent:
@HemanthTigga You've ticked the box that says the images are under 1KB. But two of the images are much larger. Why is that?

Additionally, they don't look like all the other icons, do they?

Please correct them. Ask questions if you need to.

HemanthTigga:
How do i scale down viewbox because if i use 0 0 512 512, the image just overflows. How do i fix it?

edent:
You can try using [Yann Armelin's SVG Path Editor](https://yqnn.github.io/svg-path-editor/), [svgo](https://github.com/svg/svgo), and [svgcleaner](https://github.com/RazrFalcon/svgcleaner).

HemanthTigga:
Thanks! I'll try these right away.

HemanthTigga:
updated the viewbox of all three files

edent:
@HemanthTigga Have you read the Contributing guide?

* gitBook is 1.15 KB
* mdBook won't fit in the guidelines
* mkDocs is 2.05 KB and doesn't look like their official logo

HemanthTigga:
I tried reducing the size but that's as far as I could go.

edent:
Take a look at the viewbox of gitBook - it's set to 6500! Reduce that and scale the image. That should help.

Perhaps delete the others and come back when you understand a bit more about what you're doing?

edent:
@HemanthTigga Any updates?

PR #831, opened 2024-10-04
Add Wayland logo

Zynthasius39:

```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```


Eiim:
This looks great! I'll make sure @edent is good with it since it's a bit simplified, but I think it's quite reasonable.

edent:
This is brilliant! Thanks @Zynthasius39

PR #844, opened 2024-10-15
Leetcode Logo

kycilius:

* [X] The layout of the SVG looks like this including newlines
```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```


edent:
@kycilius this is *very* nearly ready to be accepted!

Take a look at the start of your SVG.  It looks like this:

```svg
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" aria-label="LeetCode Logo">
```

Now take a look at the message you confirmed above.

Do you see the difference?

Your logo needs to be resized for a 512 viewbox. It also needs a white background.

PR #848, opened 2024-10-17
added tailwind css svg 

Hit1000:

```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```


edent:
@Hit1000 This looks really good!

Could you please format the SVG code to look like the others in the repository?

I wonder if you can make it even smaller by re-using one of the paths?

Hit1000:
@edent thanks you for your review 
now check this updates
as you had suggested its size is now 537 bytes instead of 1001 bytes

Issue #857, opened 2024-10-19
Add Playstation Logo


kycilius:
## The icon I want is:

* Name: Playstation
* URL: https://www.playstation.com/en-in/
* Link to icon: https://upload.wikimedia.org/wikipedia/commons/thumb/0/00/PlayStation_logo.svg/300px-PlayStation_logo.svg.png

kycilius:
@edent I've read through the Contributing guide and I've committed the changes to my forked repo. Should I go ahead and send a PR? Is that okay

kycilius:
For reference, here’s the link to my forked repository:
https://github.com/kycilius/SuperTinyIcons/blob/72ff69e464935a7b93016b6dd71e13458574f111/images/svg/playstation.svg

edent:
Fixed in #860

PR #860, opened 2024-10-21
Adding Playstation Logo

kycilius:

```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```

Add Playstation Logo #857


edent:
Nearly there.

* Fix the formatting so it doesn't have spaces at the start of lines.
* Name should just be "PlayStation" - no need for the word "logo".
* Looking on the official site, I think the colour should be `#0070d1`

kycilius:
Thanks for the review @edent 
I hope this will be perfect.

kycilius:
@edent is it good

edent:
@kycilius Almost!

Yours starts like this:

```svg
  aria-label="PlayStation" role="img"
  viewBox="0 0 512 512">
```

Can you see how that's different from all the other images in the repo?

Issue #875, opened 2024-10-29
Add Visa Icon


ohmschrodinger:
## The icon I want is:

* Name: Visa
* URL: https://www.visa.co.in/
* Link to icon:  https://commons.wikimedia.org/wiki/File:Visa_Inc._logo.svg

* Vector version of the logo:  https://commons.wikimedia.org/wiki/File:Visa_Inc._logo.svg /  https://github.com/ohmschrodinger/Hacktober-Fest-2024/blob/main/visa.svg

princekumar0509:
@ohmschrodinger Can you assign me this work.

edent:
@princekumar0509 have you read https://github.com/edent/SuperTinyIcons/blob/master/CONTRIBUTING.md ?

PR #876, opened 2024-10-29
Visa Logo

ohmschrodinger:

```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```

#875


edent:
I don't think the orange swoosh is used any more - see https://www.merchantsignage.visa.com/brand_guidelines

Don't forget to run it through SVG OMG or similar to shrink the paths.

ohmschrodinger:
Done!

edent:
Please can you replace the reference image with the one from the brand guidelines website. Also, replace the URl with the brand guidelines one.

The point of the reference image is to see how our SVG compares to the *official* logo.

ohmschrodinger:
Sure thing!

PR #877, opened 2024-10-30
Attempt at GeeksForGeeks logo 

oOXxTNTxXOo:

```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```


oOXxTNTxXOo:
this was my first attempt at creating a svg. hope i did ok

Eiim:
This is a great start! Just a couple comments. First, `fill="none"` instead of `fill="#ffffff00"` is shorter and even slightly better-supported. Second, you have enough small deviations from the original logo to be noticable. Fortunately, it looks pretty easy to get a lot closer. Below is an overlay of your version in purple and the original in green.
image
Moving the left circle left and right circle right by .5px:
image
Moving top points of polygons down 2px:
image
Moving top of line down 3px:
image

I don't know that those were the optimal changes, but it looks a lot better. I haven't tried implementing this in an efficient way, just playing around in Inkscape. You might also consider combining your two polygons into one `<path>`. There might be more clever `<path>` options for the green parts as well, but it would certainly be harder.

Oh, and if you can remove the last line break so the `</svg>` is on the third line, that'd be great!

PR #884, opened 2025-01-01
Fix and reduce playstation.svg icon

chriskirknielsen:
The stem for the P in the PlayStation logo was too short (see my wonderful annotated comparison below!), so I fixed it — some other changes like the unbalanced opening in the lower section of the S were also fixed. I also increased the logo size to fit the edges of the guideline circle, making the logo occupy more space. Finally, I reduced the filesize down from 703 to 515 bytes.

I based all these changes on the official logo, so it's overall a closer match while reducing filesize — hopefully this is a good improvement all around!

annotated comparison with the old and new logo, showing the changes on the old version


Eiim:
Great work, thanks!

PR #885, opened 2025-01-22
Add dolibarr

romain-dartigues:

* [ ] The layout of the SVG looks like this including newlines
```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```


images/svg/dolibarr.svg
edent, line 1: This looks too tall to me.

romain-dartigues: I'll try again, thank you for the feedback.

It looks too tall comparing to the guidelines or the original?
Here are overlay with both:
guidelines
diff

romain-dartigues:
I made a new attempt by reducing the D̊  in order to fit totally in the yellow zone and by using the more consistent.

PR #886, opened 2025-01-22
Add plausible

romain-dartigues:

```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```

Note:
This use the new (2024) logo from https://github.com/plausible/docs/pull/471.


images/svg/plausible.svg
Eiim, line 1: The SVG should mostly be on one line, but that should be easy to fix.

romain-dartigues: Ah, is it the new trend now?
@edent < sorry to bother you (^_^)
Is it the new recommandation to online the SVG, or should we try to make the diff easier?

(same question for my others contributions, like https://github.com/edent/SuperTinyIcons/pull/888#discussion_r1928142987).

romain-dartigues:
Added the `.url` ('^_^)

images/svg/plausible.svg
edent: No need for everything to be on one line.

PR #888, opened 2025-01-23
Add GraphQL

### LINKED ISSUE #828: GraphQL Logo
## The icon I want is:

* Name:  GraphQL Logo
* URL: https://cdn.freelogovectors.net/wp-content/uploads/2021/01/graphql-logo-freelogovectors.net_.png

@edent  i want to add this can you assign to this me with label  hactoberfest :)

romain-dartigues:
Reference from: https://graphql.org/_next/static/media/logo.ad338028.svg

```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```

Notes:
The logo might seems smaller than usual but I tried to respect the directive "Keep a clear space of at least half the height of the hexagraph along on all sides"[^1].

If accepted, this should close #828.

[^1]: https://graphql.org/brand/#spacing


images/svg/graphql.svg
Eiim, line 1: This one should have the SVG code compacted down to fewer lines as well.

romain-dartigues: Hi @Eiim,
Sorry but I don't understand, would you be so kind as to explain it to me a bit more?
I used new lines instead of spaces in order to try to make diffs easiers, if this is no longer desired, I could stop doing that, but the resulting file size will be the same :^)

PR #900, opened 2025-04-20
Importable flutter package (git submodule recommened) for 258 support…

gleatd01:
This added all available svg's from this repo to a flutter icon set. Should be able to import supertinyicons.dart and import the icons directly. I plan to use this directly as a git submodule in production. Also plan to break fonts/SuperTinyIcons.ttf into individual icons to make maintenance easier. This should not impact anything you currently have implemented. Did update .gitignore to prevent flutter module uploads.


pubspec.yaml
```diff
@@ -0,0 +1,58 @@
+name: supertinyicons
+description: "A new Flutter package project."
```
edent, line 2: Please add a better description here.

lib/supertinyicons.dart
edent, line 1: Can this file be autogenerated?

fonts/SuperTinyIcons.ttf
edent, line 1: How is this file created?

config.json
edent, line 1: What's the purpose of this file?

edent:
This is a great start! Thank you. I've added a few comments.

My main concern is that there's an easy way to generate these files every time a new icon is added.

pubspec.yaml
gleatd01: Can do.

lib/supertinyicons.dart
gleatd01: If I can figure out how to generate a .ttf file automatically then yes.

fonts/SuperTinyIcons.ttf
gleatd01: Manually uploaded all svg’s to fluttericon.com and deleting the svg’s that cannot be converted. Definitely interested in alternatives.

config.json
gleatd01: This might be an artifact that can be deleted, I’ll check.

lib/supertinyicons.dart
romain-dartigues: I suggest https://github.com/fonttools/fonttools

Issue #895, opened 2025-01-26
Request to automate the `update_readme`


romain-dartigues:
Rationale

It is counterproductive to run `update_readme`, and tracking in the repository files going to conflict on the next PR (CHECK.html, README.md, REFERENCE.md).

How would you feel automating the process?

A possible workflow could be something like:

```
user submit Pull Request   <------------\
|                                       |
`-- GitHub: tests ---|fail, try again|--/
    |
 success!
    |
    `-- Project administrator merge
        |
        `-- GitHub: build, publish
```

The Build process would:

1. generate websites pages (ie: "What's Available so far?", "Reference Images")
2. generate libraries[^1]

Nice to have:
* stats on the "website"
* RSS for new icons

If you are interested, I made a minimal proof of concept on:
https://github.com/romain-dartigues/SuperTinyIcons/tree/github-pages
automatically publishing the website at:
https://romain-dartigues.github.io/SuperTinyIcons/index.html

It's quite ugly for now, I'm confident you'll be able to give it a better look :^)

[^1]: I would like to generate [Draw.io](https://draw.io/) icons libraries, [texture atlas](https://en.wikipedia.org/wiki/Texture_atlas) a.k.a. spritesheet, downloadable archive…
I will open issue and provide PR on this topic later if you show some interest.

PR #902, opened 2025-06-15
Add victoria metrics

romain-dartigues:
```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```


edent:
Looks good. Could you please add a link (using a .url file) so it appears in the reference doc. Ta!

romain-dartigues:
You are alive! glad to hear about you :^D

Your comments should have been addressed.
The failing test seems to be related to hammering the W3C validator.

Did you get a chance to look at #895 and it's links?

edent:
:-) I've been on a lovely long holiday. Thanks for the icon!

PR #912, opened 2025-10-01
Add NordVPN icon and reference assets

SaralDwivedi21:

```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```

Please add this as hacktoberfest contribution


edent:
This is close. You don't need the title element. Could you also give it a square background rather than a circle?

Thanks.

SaralDwivedi21:
Hi @edent 
I have updated the svg. Could you please check if it's alright.


I also have a new svg that i would like to add but don't know if i need to open a new pr or commit it to this one. I would be really grateful if you could help me with that since i am relatively new to open source as a whole.

edent:
@SaralDwivedi21 it should be a new PR.

PR #940, opened 2025-10-03
add appwrite logo

fadkeabhi:

```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```

image
Some part of icon is kept in yellow circle as keeping all in green is not visually appealing.
because 
1. either it is small in size.
image

2. icon biased to left side
image


images/svg/appwrite.svg
```diff
@@ -0,0 +1,9 @@
+<svg xmlns="http://www.w3.org/2000/svg"
+aria-label="appwrite" role="img" 
+viewBox="0 0 512 512">
+<path d="m0 0H512V512H0" fill="#fff"/>
```
edent: You can remove *some* of these newlines to reduce the filesize.

images/svg/appwrite.svg
```diff
@@ -0,0 +1,9 @@
...
+  <g id="logo" transform="translate(86 106) matrix(3 0 0 3 0 0)">
```
edent: Use SVGOMG or similar to remove the need for this transform.

edent:
I think you've made the right choice for size / placement. I've left a couple of comments - if you can fix those, I'll happily merge it.
Thanks!

images/svg/appwrite.svg
fadkeabhi: updated with format similar to other svg files.

images/svg/appwrite.svg
fadkeabhi: Used SVGOMG to remove transform

fadkeabhi:
Hi @edent, I've fixed the issues as per your comments. Please take a look.

Issue #801, opened 2024-09-10
Missing reference images & style guides


edent:
As per: https://github.com/edent/SuperTinyIcons/blob/master/REFERENCE.md#no-reference-image-found

There are several images which have no reference image and no style guide associated with them.

To fix this, you need to add two files.

1. `servicename.png` the official logo downloaded from their website. Can be in .png, .svg, or .jpg format.
2. `servicename.url` a text file which contains a single line; the URl of the *official* brand guidelines / style guide.

Note: `servicename` *must* match the name of the icon in `/images/svg/`.

Place both of those in `/images/reference/` and then run `python update_readme.py`

justynigam:
I love to do this issue , **Please assign me** 
I can provide in a good way.

edent:
@justynigam you do not need to be assigned. Please just go ahead. Remember to follow the contributing guide.

SaurabhKumar171:
I would like to contribute to this issue. Should I go ahead and start working on it, or will it be assigned to me first?

edent:
@SaurabhKumar171 have you read https://github.com/edent/SuperTinyIcons/blob/master/CONTRIBUTING.md ?

SaurabhKumar171:
@edent yeah , got it. Thanks!

PR #941, opened 2025-10-03
feat: add missing reference images and style guides

conventoangelo:
Thank you for your contribution! Before sending this Pull Request, please confirm the following:

---

I tried my best in finding the brand guidelines but defaulted to some other sources when the service is old. This PR tries to solve issue #801.

Note that some references already exist in the repository but with an incorrect name. This were either renamed only or improved upon.


edent:
Brilliant work - thanks!

Issue #757, opened 2023-09-01
Fedora Logo Request


solidc0re:
## The icon I want is:

* Name: Fedora
* URL: https://fedoraproject.org/
* Link to icon: https://fedoraproject.org/wiki/File:Fedora_infinity.png

* Vector version of the logo: https://fedoraproject.org/wiki/Logo/UsageGuidelines#Obtaining_the_Fedora_logo_for_use_by_graphic_designers
* Official style guide or brand guide: https://fedoraproject.org/wiki/Logo/UsageGuidelines

kuubeu:
The guidelines mention that [the logos should not be replicated](https://fedoraproject.org/wiki/Logo/UsageGuidelines#Don't_attempt_to_replicate_the_logos). Should I add the icon regardless?

nishitiwari22:
Hello, I am looking forward to contributing to this. I have a small question regarding this You want to have an icon like Fedora but not Fedora. I mean I can take inspiration from the logo but the logo design should be different. Am I right?

Eiim:
No, I don't think we want logos that aren't official logos. It's not clear whether or not we should add this logo (as that would involve replicating it).

Eiim:
That said, we do have #698 - that was a bit of an exceptional circumstance as there was no way the full logo could fit in 1kb, but this is also a bit of an exceptional circumstance. But there is some precedent for "not-real" logos. Maybe using the [old logo](https://commons.wikimedia.org/wiki/File:Fedora_icon_(2021).svg) would be less problematic?

NitinSingh07:
Sir, Can you please assign this work to me?

Munazir151:
can I work on this?

Potherca:
Regarding the brand guidelines, I asked for clarification over on the the [Design team’s GitLab](https://gitlab.com/fedora/design/team).

Follow https://forge.fedoraproject.org/design/tickets/issues/29 for a resolution (if any).

PR #955, opened 2025-10-04
Fedora Logo SVG

Debatreya:

```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```

#757


Debatreya:
@edent I have tried to put the Fedora logo as per the issue , in this PR.
(The PR may show a line of other commits from my Flask work, I dont know why, But you can check the files changed to be sure that only Fedora related changes are there.  I can try fixing that, if you want this fixed)

Please check the svg once, I have tried to keep the format same this time. Suggest changes , I am happy to fix them.

edent:
This is close, but it doesn't quite fit in with the others. I suggest making the width of the image 512 rather than leaving the white border.

Debatreya:
> This is close, but it doesn't quite fit in with the others. I suggest making the width of the image 512 rather than leaving the white border.

Okay I got your point. I will make the changes

Debatreya:
@edent Can you review it now.

edent:
Can you use SVGOMG to reduce the size while keeping the quality?

Debatreya:
> Can you use SVGOMG to reduce the size while keeping the quality?

@edent Yeeah did this thing, its now about 465 Bytes in size

Issue #952, opened 2025-10-04
[Request] add McAfee logo


Muteeb489:
The icon I want is: 

* Name: McAfee
* Website: https://www.mcafee.com/consumer/en-us/landing-page/direct/aff/mtp-family/desktop/mcafee-total-protection.html?culture=en-us&SID=cbee87f53f11452e84a28505668e326c16e7d&cjevent=ae9bd0baa10e11f0812101670a82b836&affid=1494&csrc=cj&csrcl2=DMS+CD+%28Netherlands%29+B.V&ccoe=direct&ccoel2=am&ccstype=partnerlinks_ae9bd0baa10e11f0812101670a82b836&CID=242012&PID=101096380&prgt=lc&gad_source=1&gad_campaignid=22637366122&gbraid=0AAAAA_tzVCvBkCB4ntSyDf-5e-Ej3jgbH&gclid=CjwKCAjwi4PHBhA-EiwAnjTHuRS4zPUxPKmZ9JjBp4xAVuR4SZAro3OiSYzer-zW6qhDzReOvaEB7xoCQvsQAvD_BwE
* Link to icon: https://assets.esecurityplanet.com/uploads/2024/08/mcafee-icon.png

Prakhar0013:
Hi I got the svg for this. Can I try?

edent:
@Prakhar0013 have you read https://github.com/edent/SuperTinyIcons?tab=readme-ov-file#hacktoberfest ?

Prakhar0013:
Yeah.

edent:
Well then, you know the answer!

Prakhar0013:
I have a doubt regarding pull request.

Prakhar0013:
After mistake in my last pull request, I have been checking the format of svg, trying to correct it so that it appears properly in readme. I have been trying to correct it many times but to no avail.
Please let me know my mistake. Why is my svg not appearing in readme? I created mcafee.svg, added it to the images/svg/. And I ran python3 update_readme.py(In my forked repo). But after many tries as well it didn't appear there. Size is right now, I am sure. I have even ensured that the image appears in the circle.
Please let me know my mistake.
[...]

PR #956, opened 2025-10-04
Added McAfee SVG image

Prakhar0013:

* [✔️ ] You have read [the contributing guidelines](https://github.com/edent/SuperTinyIcons/blob/master/CONTRIBUTING.md)
The file name is mcafee.svg thus with right name convention
* [✔️ ] The icon's size is *under* 1,024 Bytes
The icon size is 363 bytes(in my system) thus under the range
* [✔️ ] The layout of the SVG looks like this *including* newlines
```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```
I did a mistake of not adding aria-label earlier but corrected it later when encountered error in running update_readme.py.
Thats it. I can even explain how I got this.


Prakhar0013:
This is with regards to issue [Request] add McAfee logo #952. Please check it.

Prakhar0013:
If possible I wanted a hint in another issue please. Actually regarding W3C Label I tried a lot but only was able to make it to 13-14 kb. It contains text as you may know. I tried using inkscape to minimize the coordinates/points(Now I can't mess with many points as it may mess with quality). Still even after many optimizations  wasn't able to get it right. Can you please share an idea/hint? I have tried many SVG Optimizers(online as well as on GIthub).

edent:
We ask that the layout looks like this:

```
aria-label="..." role="img"
viewBox="0 0 512 512"><path
```

Yours looks like this:

```
<svg fill="#c8102e" width="800px" height="800px"
aria-label="McAfee" viewBox="0 0 24 24" role="img" 
```

As per the Guidance (which you said that you read) https://github.com/edent/SuperTinyIcons/blob/master/CONTRIBUTING.md#guidelines - your image must fit in the green zone of this image:

A template for logos

This isn't even close to being suitable.

You need to *read* the guidance. You also need to *follow* the guidance.

Prakhar0013:
Sorry, it was my mistake. Will be careful.

PR #960, opened 2025-10-05
Added obs logo

blezecon:
I have followed all the contribution guidelines:

✅ I have read the [contributing guidelines](https://github.com/edent/SuperTinyIcons/blob/master/CONTRIBUTING.md)

✅ The icon’s file size is 961 Bytes.

✅ The SVG layout follows the required format, including newlines, as shown in the example.

Thank you for reviewing my contribution!
Screenshot_20251004_104338


edent:
I think there's a couple of ways to shrink this further.

Take a look at how many decimal places the path co-ordinates have. Do you lose anything if you round them?

Could you make just one "spike" and then rotate it for the others?

blezecon:
Current Size: 741B

PR #964, opened 2025-10-05
Add Google Pay icon

reginareynolds:

```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```


images/svg/google_pay_old.svg
```diff
@@ -0,0 +1,5 @@
+<svg xmlns="http://www.w3.org/2000/svg"
+aria-label="GPay" role="img"
+viewBox="0 0 512 512"><path
+d="m0 0H512V512H0"
+fill="#fff"/><g transform="translate(64,64) scale(8)"><path fill="#e64a19" d="M42.9 12c-4.5-2.6-10.4-1.1-13 3.5L23.3 26.9c-1.9 3.3.6 4.5 3.3 6.1l6.4 3.7c2.2 1.2 4.9.5 6.2-1.6L46 23.3c2.3-4 .9-9-3-11.3z"/><path fill="#fbc02d" d="m35.4 16.7L29 13c-3.5-2-5.5-2.1-7 .2l-9.4 16.3c-2.6 4.5-1.1 10.4 3.5 13 4 2.3 9 .9 11.3-3L37 22.9c1.3-2.2.5-4.9-1.6-6.2z"/><path fill="#43a047" d="m36.6 8.4-4.5-2.6c-5-2.9-11.3-1.2-14.1 3.8L9.5 24.3c-1.3 2.2-.5 5 1.7 6.2l5 2.9c2.5 1.4 5.6.6 7.1-1.9l9.7-16.9c2-3.5 6.5-4.7 10-2.7L36.6 8.4z"/><path fill="#1e88e5" d="M19.2 13.8c2.2 1.2 2.9 4 1.7 6.1l-5.5 9.5-1.5 2.6c-2.1 3.6-1 8.3 2.4 10.6l-1.9-1.1-4.5-2.6-3.7-2.1c-5-2.9-6.7-9.2-3.8-14.1L8.2 12.6c1.2-2.2 4-2.9 6.2-1.6l4.8 2.8Z"/></g></svg>
```
edent, line 5: Use SVGOMG to apply the transformation and scale. Remember, try to get it as small as possible.

images/reference/google_pay_old.url
```diff
@@ -0,0 +1 @@
+https://about.google/brand-resource-center/logos-list/
```
edent, line 1: Can you please add a reference image as well?

reginareynolds: The contribution guidelines said to use an official image, and since this is the old Google Pay icon, all the current brand guidelines have assets for the new Google Pay icon. After poking around, the Google logos list still shows it as an official registered icon, but the only reference I can find to it is the little image on the logos list, which is in WEBP image format. As far as I can tell, all of the reference photos so far are either PNG, JPEG, or SVG format. Should I just upload the WEBP image as the reference image?

images/svg/google_pay_old.svg
reginareynolds: It's actually smaller not to bake the transformation and scale into the SVG coordinates. I tested both ways before submitting the smaller one. By keeping the translation and scale unapplied, the file size is 834 bytes, but baking it in brings it to 880 bytes.

images/reference/google_pay_old.url
edent: Webp is fine. But if this is the old logo, I don't think there's any point us having it. We only have the old Google icons because they were new when we started.

reginareynolds: Alright, I'll see about shrinking the new Google Pay icon to meet the size requirements. Given that the old icon is still floating around in the wild at POS registers, I'll rename the old icon and reference files to indicate that it represents the old version while I get the new version within the size constraints.

Issue #878, opened 2024-10-31
Use fill="none" for fill="#fff" for background color


shinokada:
These days developers use dark/light/system modes. If you have `fill="#fff"`, you have the white bg when the mode is dark.

For example, bluesky has it.

I recommend using `none` for `#fff`.

edent:
That's an interesting idea. Could you please show a screenshot comparing them?

shinokada:
The following screenshots are on the dark mode.

With `fill="none"`
![image](https://github.com/user-attachments/assets/86b97b34-2706-4bbf-ae90-1631678d959f)

With `fill="#fff"`
![image](https://github.com/user-attachments/assets/0c7b3b56-0495-461d-b8e5-0cdcf5361411)

edent:
Looks good. Are there any icons where the background colour forms part of the logo?

shinokada:
There are many SVG icons have `<path d="m0 0H512V512H0" fill="#fff" />`.

Muteeb489:
Hey if you want I can write a script tag to manually go over the svg and trigger fill or no fill depending on the user background. Sound good?

Eiim:
Per @edent's earlier question, there are definitely some logos that this would cause issues with. Consider just the second icon in our set alphabetically: Acast would be basically invisible on a dark background, and probably would need an alternate fill, depending on how the original brand does it.

Talhaasif7:
Can you guys please review my PR.
Lemme know if it still have any problem.

edent:
@Talhaasif7 We did review it. Did you read our comments?

PR #965, opened 2025-10-05
fix[#878]: make all icon backgrounds transparent for dark mode support

Talhaasif7:
### Summary
This PR fixes dark mode issues by replacing all icon background paths using `fill="#fff"` (white background) with `fill="none"`.  
This makes icons display correctly in dark, light, and system themes.

### What I Changed
- Updated all `/images/svg/*.svg` files that used a white background to `fill="none"`.
- Regenerated `README.md`, `REFERENCE.md`, and `CHECK.html` with `python3 update_readme.py`.
- Verified file sizes remain under 1KB.
- Ran `npm test` successfully.

### Testing
All icons display properly with transparent backgrounds.

Fixes: #878

---

### ✅ Confirmation

* [x] The icons are each *under* 1,024 Bytes  
* [x] The SVG layout follows the required format:
  ```svg
  aria-label="..." role="img"
  viewBox="0 0 512 512"><path
  d="m0 0H512V512H0"
  fill="none"/> ... </svg>
  ```


edent:
Have you tested what the images look like on a black background? For example, what does the Auth0 one look like?

Eiim:
I'm in opposition to this. Some icons simply can't work this way in dark mode: acast, angellist, and apple are obvious cases from a quick glance at the icons in this PR. For some, the white is used as part of the icon and should not be replaced with a different color: [bash](https://github.com/odb/official-bash-logo/blob/master/assets/bash-style-guide.pdf) is a great example. Some icons are already exclusively in dark mode: buffer and codepen are two such cases. Some icons change their own colorways in dark mode: again, bash is an example. If we want to support separate dark and light modes, we'd need to take a more holistic look at it.

PR #968, opened 2025-10-06
Added Linux foundation Logo

blezecon:
I have followed all the contribution guidelines:

✅ I have read the [contributing guidelines](https://github.com/edent/SuperTinyIcons/blob/master/CONTRIBUTING.md)

✅ The icon’s file size is under 1,024 Bytes.

✅ The SVG layout follows the required format, including newlines, as shown in the example.

Thank you for reviewing my contribution!


edent:
* This doesn't fit withing the circle.
* You also need to run `python update_readme.py`

blezecon:
Screenshot_20251006_123748
used the microsoft svg for comparing, because the guideline one is circle.

images/svg/linux-foundation.svg
edent, line 1: This doesn't look like the reference image. The gaps are too large.

blezecon: yea I know, the reference one is circle, thats why I used this as reference, the microsoft one is already in the repo.

blezecon: trying to fix it rn

blezecon:
Screenshot_20251006_134122
its a square image, so some part got in the red area, want me to make the logo smaller, or this size is ok. @edent

edent:
You don't need to @ me each time. GitHub sends me notifications.

Size is OK - but make sure it is centred.

edent:
This isn't good enough. Here's what the images look like when displayed as an animation:

Animation showing different images.

The point of the reference image is that the SuperTinyIcon *must* be as close as possible to it.

blezecon:
It's a stupid question I know. But what if I change the referance image.

Eiim:
@blezecon Don't change the reference image. The point of the reference image is to be exactly what the brand uses for their logo. This logo is simple enough that it shouldn't be hard to fix.

blezecon:
Just asking, will fix soon.

blezecon:
Screenshot_20251006_201941

PR #970, opened 2025-10-06
added people-per-hour Logo

ARP-Proworks07:

* [ ] You have read [the contributing guidelines](https://github.com/edent/SuperTinyIcons/blob/master/CONTRIBUTING.md)
* [ ] The icon's size is *under* 1,024 Bytes
* [ ] The layout of the SVG looks like this *including* newlines
```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```

proof


ARP-Proworks07:
please check and merge this under hactoberfest2025

Eiim:
It looks like the company name is PeoplePerHour, so it should be formatted as such (and the file names should be peopleperhour).

Also, you can reduce the file size somewhat by removing the extra newlines and whitespaces, using a shorter id for the gradient (like #a), and removing the `<defs></defs>` tags (which aren't necessary if you just have gradients in there).

Otherwise, good work!

Eiim:
Actually, I just realized another issue. The actual PeoplePerHour logo has the gradient going across the face and eyes as one, whereas in your version the gradient is separate for each eye. It looks like you just used your SVG for the reference SVG, which isn't correct. Try to pull an image from their website to use as a reference.

PR #973, opened 2025-10-07
Add Ente auth icon

SirMonteiro:

```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```


edent:
Does the logo you've submitted fit the style of all the other logos?

SirMonteiro:
It was missing centering it, now it fit perfectly!

Eiim:
I notice that this is missing the drop shadow on the original. It's subtle enough that it might be fine, but I'll leave that up to edent.

SirMonteiro:
I removed it to optimize even more, this original icon has gradient and shadow!

edent:
If you can add the gradient and shadow, that should still be within the size limit.

SirMonteiro:
I added linear gradient to the icon, tried to optimize using all three tools, svgo CLI, svgomq and path editor, so now the icon has 540 KiB, I tried to add shadows, but each one of them has more than 1.5 Kb, so I think it's not possible.

SirMonteiro:
I don't know why test is not passing, verifying with https://validator.w3.org/check it passed.

Eiim:
The validator wants your first stop to explicitly declare offset="0". This is optional in SVG 2, but not earlier versions (although any SVG renderer worth its salt will handle it fine anyways)

SirMonteiro:
While I agree with you, for maximum compatibility and to pass project actual validator, I added explicit offset on first stop, thank you to clarify why the test was not passing!

PR #976, opened 2025-10-08
Feat: add target logo

Krasper707:

- [x] You have read [the contributing guidelines](https://github.com/edent/SuperTinyIcons/blob/master/CONTRIBUTING.md)
- [x] The icon's size is *under* 1,024 Bytes
- [x] The layout of the SVG looks like this *including* newlines
```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```
Reference Files: The optional reference image (target.png) and official brand URL (target.url) are included.
Project Files: The `README.md` , `CHECK.html` , and `REFERENCE.md` files have been successfully updated by running the update_readme.py script.


edent:
Thanks. This doesn't really fit in with the theme of SuperTinyIcons. We're looking for Web / Internet / Tech icons.

It is technically very good - so I've given you the hacktober accept label.

If you'd like to contribute any more - please raise an issue with the icon you want so we can discuss if it is suitable.

PR #957, opened 2025-10-04
Add Notion SVG icon

reginareynolds:
Thank you for your contribution!  Before sending this Pull Request, please confirm the following:

* [X] You have read [the contributing guidelines](https://github.com/edent/SuperTinyIcons/blob/master/CONTRIBUTING.md)
* [X] The icon's size is *under* 1,024 Bytes
* [X] The layout of the SVG looks like this *including* newlines
```svg
<svg xmlns="http://www.w3.org/2000/svg"
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```

If you have done the above, please send the Pull Request.

PR #980, opened 2025-10-09
Change Notion URL to brand guidelines page.

Potherca:
The URL to the Notion Guidlines (added in #957) is incorrect.

This MR changes the URL to the correct page.


images/reference/notion.url
```diff
@@ -1 +1 @@
-https://www.notion.com/about
+https://www.notion.so/notion/Notion-Trademark-Usage-Guidelines-9826313c686a4f6e9d8a48347162714b
```
edent, line 1: What we're looking for here is a link to the page which shows how the icon should be used - no just text about trademarks. Is there a better page on Notion's webiste?

Potherca: The other available URL is: https://www.notion.so/notion/Notion-s-brand-usage-guidelines-How-to-use-Notion-s-brand-in-your-marketing-30a5510bc5644475a28844e427008bee

Both pages link to each other, so I wasn't sure which one to use...

edent: I think go for that 2nd one as it has more images on it. Thanks.

PR #982, opened 2025-10-10
Fitbit icon

reginareynolds:
Thank you for your contribution!  Before sending this Pull Request, please confirm the following:

* [X] You have read [the contributing guidelines](https://github.com/edent/SuperTinyIcons/blob/master/CONTRIBUTING.md)
* [X] The icon's size is *under* 1,024 Bytes
* [X] The layout of the SVG looks like this *including* newlines
```svg
<svg xmlns="http://www.w3.org/2000/svg"
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```

If you have done the above, please send the Pull Request.

PR #985, opened 2025-10-11
Further shrink Fitbit icon

reginareynolds:

```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```


reginareynolds:
After optimizing the path provided by @Eiim in #982  using [Yann Armelin's SVG Path Editor](https://yqnn.github.io/svg-path-editor/), it went from 504 characters to 501 characters, which ended up looking like this:
```svg
M248.3 390.5a1 1 0 000 57.6 1 1 0 000-57.6m-80.3-79a1 1 0 000 52.2 1 1 0 000-52.2m80.3-3.9a1 1 0 000 60 1 1 0 000-60m79.4-3.8a1 1 0 000 67.6 1 1 0 000-67.6M87.4 232.6a1 1 0 000 46.8 1 1 0 000-46.8m80.6-3.9a1 1 0 000 54.6 1 1 0 000-54.6m80.3-3.8a1 1 0 000 62.2 1 1 0 000-62.2m79.4-3.9a1 1 0 000 70 1 1 0 000-70m81.5-3.8a1 1 0 000 77.6 1 1 0 000-77.6M168 148.3a1 1 0 000 52.2 1 1 0 000-52.2m80.3-3.9a1 1 0 000 60 1 1 0 000-60m79.4-3.8a1 1 0 000 67.6 1 1 0 000-67.6M248.3 64a1 1 0 000 57.6 1 1 0 000-57.6
```

I replaced the contents of the original Fitbit SVG that I created with this path, and that reduced the file size from 719 bytes to 669 bytes! But then I got to thinking about the way that I had translated the circles originally and decided to take a closer look, and I realized that there were a couple things I could do to shrink it further.

For example, I originally gave the circle at the very top of the icon an id, then used a `<use>` element that referenced it and translated it to the bottom of the logo, but there were actually more characters in the `<use>` element than just explicitly defining that as a new `<circle>` element.

I also realized that since I was translating every circle in the second row by the same amount to copy it down to the fourth row, I could apply the translation to the row as a whole rather than to each circle in the row, which meant I would only have to use a single `<use>` element. Once I did that, I was able to shrink the file even more to 604 bytes!

Eiim:
Looks great, well done!

Eiim:
Before I merge though, @edent it looks like the README is affected by line endings (probably counting like Windows), is that fine to merge? Perhaps the Python script needs fixed somehow there?

edent:
Yeah, fine to merge. Might need to re-run the script manually. It should eventually become a GitHub action.

PR #986, opened 2025-10-11
Add socketIO icon

Kshitizjain11:

```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```

I couldn't find a separate brand guidelines page of the socketIO so included the socketIO website link in the url file.


images/svg/socketIO.svg
```diff
@@ -0,0 +1,4 @@
+<svg xmlns="http://www.w3.org/2000/svg"
+aria-label="Socket.IO" role="img"
+viewBox="0 0 512 512"><path d="m0 0h512v512h-512z"
+fill="#fff"/><circle cx="256" cy="256" r="228" fill="none" stroke="#010101" stroke-width="40"/><path fill="#010101" d="M195.274 243.38c54.654-44.652 108.116-90.852 163.96-134.194-29.292 45.01-59.416 89.422-88.708 134.43-25.124.12-50.246.12-75.252-.236zm46.2 24.884c25.242 0 50.366 0 75.49.358-55.01 44.412-108.234 90.968-164.198 134.192 29.292-45.01 59.416-89.54 88.708-134.55z"/></svg>
```
edent: Try using SVGOMG to reduce this. You don't need 3 decimal places of precision.

Kshitizjain11: Hi @edent,
I’ve reduced the SVG precision to zero and made a few fixes. I was trying branch relate stuff due to which this PR got closed. Also i guess, reducing precision may reduce the size of svg but in this case it didn't result into showing of that socketio svg

PR #981, opened 2025-10-09
Add socket.io icon

GlyphicGuy:
Thank you for your contribution!  Before sending this Pull Request, please confirm the following:

* [X] You have read [the contributing guidelines](https://github.com/edent/SuperTinyIcons/blob/master/CONTRIBUTING.md)
* [X] The icon's size is *under* 1,024 Bytes
* [X] The layout of the SVG looks like this *including* newlines
```svg
<svg xmlns="http://www.w3.org/2000/svg"
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```

If you have done the above, please send the Pull Request.
Fixes #977 
**Clarificaiton**:
The icon is complete white on black background in the dark theme, should we do something about that too?
Kindly let me know and I shall include that too.
These are dark and light theme logos respectively (these are mere screenshots from the website):

image 
image

PR #988, opened 2025-10-12
reduced precision to zero for SocketIO icon

Kshitizjain11:

```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```

I made changed as said but it didn't shown as earlier


Kshitizjain11:
i @edent ,
I’ve reduced the SVG precision to zero and made a few fixes. Earlier, I realized the file names need to be in lowercase — my previous version used socketIO, which might have caused an issue. I’ve corrected that in this PR.

I also tried experimenting with some branch-related changes earlier, which may be why the previous PR was closed. However, I’m still unable to fix that preview error completely yet.

images/reference/socketio.svg
edent, line 1: The reference image needs to come from the original site. That way we can compare your SuperTinyIcon and the official one.

edent:
Closed in favour of  #981

PR #989, opened 2025-10-12
Create the Arduino Icon

mxritzdev:

```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```


images/svg/arduino.svg
```diff
@@ -0,0 +1,7 @@
+<svg xmlns="http://www.w3.org/2000/svg"
+aria-label="arduino" role="img"
+viewBox="0 0 512 512">
+ <path fill="#00979d" d="M449.01,255.91c0,106.6-86.42,193.01-193,193.01-106.61,0-193.03-86.41-193.03-193.01S149.4,62.92,256.01,62.92c106.58,0,193,86.42,193,193h0Z"/>
```
edent: This is an excellent start. See if you can reduce the precision - for example does M449.01 need that .01?

PR #990, opened 2025-10-13
Add Google Wallet icon

reginareynolds:

```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```


Eiim:
Can you check this against the guideline circles? I think it's probably too big.

edent:
It also needs to be vertically centred. At the moment it is slightly too high up. Looks good though!

reginareynolds:
I fixed the scaling and centering!

Eiim:
Looks good, thanks!

Issue #991, opened 2025-10-13
Replace Tutanota


edent:
Tutanota has been rebranded to Tuta.

https://mail.tutanota.com/login

https://tuta.com/

Debatreya:
Hey, Trued to resolve this in PR - #992 
Added the Tuta logo (before hover state) If you do to the site, it actually toggles on hover.

And didn't do anything about the 'Replace' thing, does it mean Should I also delete the tutanota.svg ?

edent:
No, I think it is OK to keep the old one. Have left a comment on your file - once it is centred, I'll merge it. Thanks!

PR #992, opened 2025-10-14
feat: Tuta logo Added

Debatreya:

```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```

Solving Issue - #991


images/svg/tuta.svg
edent, line 1: This looks good - could you centre it please? It looks a little offset to the left.

Debatreya: Can you review it once , I have tried to align it to the middle

PR #994, opened 2025-10-14
Riot

arthur00007:

* [ ] You have read [the contributing guidelines](https://github.com/edent/SuperTinyIcons/blob/master/CONTRIBUTING.md)
* [ ] The icon's size is *under* 1,024 Bytes
* [ ] The layout of the SVG looks like this *including* newlines
```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```


Eiim:
This could be a lot better optimized, I would suggest avoiding transformations and reducing precision. But I'm not sure that game publishers are a category we want to cover in the first place.

arthur00007:
I have done what you said. I have avoided using transformations and also reduced precision.

edent:
You're getting there. The precision has been reduced, but now it looks nothing like the original image.

See if you can shrink the image filesize while keeping it looking good.

arthur00007:
I have fixed the shape of the fist; now it looks like the original logo.

Eiim:
I think just one thing: the aria-label should be capitalized, and then you need to re-run the Python script to get the correct name and file size. Once that's done, it should be good to merge.

arthur00007:
I have capitalized the aria-label and also re-run the Python script.

PR #1002, opened 2025-10-17
Adding Google's Find Hub icon

piyush28verma:
```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```

Adding Google's Find Hub icon

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,6 @@
 # Super Tiny Icons
 
-<<<<<<< HEAD
 Under 1KB each! Super Tiny Web Icons are minuscule SVG versions of your favourite logos. There are currently 444 icons and the average size is _under_ 524 bytes!
-=======
-Under 1KB each! Super Tiny Web Icons are minuscule SVG versions of your favourite logos. There are currently 444 icons and the average size is _under_ 524 bytes!
->>>>>>> upstream/master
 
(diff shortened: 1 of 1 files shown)


Eiim:
Make sure to read the contributing guidelines. Icons should not fill up the full square and should have a background. I would also recommend trying to incorporate the transforms into the path data.

piyush28verma:
Thanks for your feedback, I have tried incorporating required changes in this commit.

Eiim:
Much better, but it still extends into the red zone. I would recommend keeping this entirely within the green.

piyush28verma:
I hope it is fine now.

edent:
Use a tool like https://svgomg.net/ to get rid of the transform. That will shrink the size considerably. Don't forget to re-add the linebreaks.

piyush28verma:
Thanks for guidance! Please have a look

edent:
Getting closer! The colours seem a little off. Use a colour picker tool on the original image and make sure the SVG uses the *exact* colours.

piyush28verma:
I have used colour picker tool on official icon and fixed the off colours.

edent:
Excellent work - thanks :-)

piyush28verma:
Thanks to you both too for guiding me at each step 😇. I got to learn a lot through this pr!

PR #1010, opened 2025-10-20
Added Express js icon

Kshitizjain11:

```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```

Added Express js icon


images/reference/expressjs.url
```diff
@@ -0,0 +1 @@
+https://expressjs.com/en/resources/community.html/
```
edent: This 404s

Kshitizjain11:
Done. please check
Last ```/``` was not needed. I updated it

PR #1016, opened 2025-10-20
Adding SVGOMG icon

piyush28verma:
* [ ] I have run `python check.py` and received a ✅
```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```

Adding SVGOMG icon.


piyush28verma:
I tried running  python3 update_readme.py and new command : python check.py but both didn't work.

PR #1026, opened 2025-10-21
Added Upstash Icon

Kshitizjain11:

```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```
Added Upstash Icon


Eiim:
A few notes:

* You need a background fill.
* Why are the light areas implemented as two paths per part, one with transparency? If you're going the transparency route, have a large transparency area that doesn't need much path data with a clip path to only the part of the logo. Otherwise, just set the color of the light regions to the intended shade.
* The colors you have don't match either the light mode or dark mode colors of the logo, based on the brand page as well as your reference image.
* Regions of the same color can be merged to a single path to save on bytes.

Kshitizjain11:
#### Thanku Sir.
I have now used color picker for the colors of the logo and grouped the same color paths.
Is it fine now

PR #1066, opened 2025-10-25
Photoshop

Durgeshwar-AI:

```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```


edent:
Did you *really* run `python check.py`?

Durgeshwar-AI:
> Did you _really_ run `python check.py`?

Yeah I did and corrected the svg according to the warnings but didn't check again

edent:
Please check again. As you can see, this doesn't match the format of the other icons.

Durgeshwar-AI:
> Please check again. As you can see, this doesn't match the format of the other icons.

Done now it has passed all the checks for the check.py

edent:
This is *very* close. You need to make the curves smoother so that they match the original. At the moment, yours looks like:
angular lines.

Durgeshwar-AI:
> This is _very_ close. You need to make the curves smoother so that they match the original. At the moment, yours looks like: angular lines.

Please check now

image

edent:
You're getting very close! I think you need a bit more precision on the letters.
Here's what the original and yours look like - can you see the difference?
photoshop

Durgeshwar-AI:
Screenshot 2025-10-29 152921
Please veriffy it now.

images/svg/photoshop.svg
edent, line 1: OK, there are a few simple things you can do to shrink this. Do all the path's need IDs? No - get rid of them.

Rather than using a style, it's probable shorter to put the elements in a group.

The background colour should use the same template as all the other images in the repo.

Finally, is 2 decimal places of precision overkill? What happens if you reduce it to 1 or 0?

Durgeshwar-AI:
Made code changes as per the request

Kept precision to 1 decimal points.

PR #1081, opened 2025-10-30
Add WorkOS

DragonSenseiGuy:

```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```

Please also add the hacktober tag.
Also I would appreciate a fast response as hacktober is ending.
Thanks for understanding!


images/svg/workos.svg
```diff
@@ -0,0 +1,13 @@
+<svg
+aria-label="WorkOS logo"
+viewBox="0 0 512 512"
+xmlns="http://www.w3.org/2000/svg"
+width="1080"
```
edent: Does this match our existing style?

images/svg/workos.svg
```diff
@@ -0,0 +1,13 @@
...
+height="1080"
+role="img"
+>
```
edent: Why are there these extra line breaks here?

images/svg/workos.svg
```diff
@@ -0,0 +1,13 @@
...
+  <g fill="#6363f1">
+    <path d="M61.495 238.136c0 7.81 2.13 14.907 5.681 21.295l68.884 119.255c7.1 12.065 17.753 22.003 31.246 26.264 25.563 8.518 53.259-2.13 66.752-24.845l16.33-29.103-65.33-113.574 69.594-119.964 16.332-28.394q7.456-12.777 19.175-21.295H183.636c-18.46 0-36.216 9.938-45.446 26.264L67.175 216.841c-3.549 6.39-5.68 13.487-5.68 21.295"/>
```
edent: Is this path optimised? Are 3 decimal places necessary?

images/svg/workos.svg
```diff
@@ -0,0 +1,13 @@
+<svg
+aria-label="WorkOS logo"
```
edent: This doesn't match our other icons.

edent (changes requested):

jmb:
(Moved my comments here)
This is very close but does not match the requested file style and does not fit fully in the guideline green circle. Please double check the guidelines.

You might be able to reduce the file size by using a clone of one path and rotating by 180 degrees (`<use href="#a" transform="rotate(180, 256, 256)/>`)

Also the photoshop.svg file should not be in this PR. Please try again!

DragonSenseiGuy:
> Also the photoshop.svg file should not be in this PR. Please try again!

Seems to be an accident, though I still don't know why the `python check.py` kept changing the file. I remember it changing the file but, I forgot to not `git add` it.
Thanks!

images/svg/workos.svg
DragonSenseiGuy: Fixed I think. Can you take a look?

images/svg/workos.svg
DragonSenseiGuy: Fixed I think. Can you take a look?

images/svg/workos.svg
DragonSenseiGuy: Removed the height. Also changed the role place.

images/svg/workos.svg
DragonSenseiGuy: I don't really have experience with optimization. Can you tell me how? Thanks!

edent:
@DragonSenseiGuy you need to read https://github.com/edent/SuperTinyIcons/blob/master/CONTRIBUTING.md#shrinking

You ticked the box saying that you had read it.

The reason for that is to help you understand what we require for this project.

DragonSenseiGuy:
> You ticked the box saying that you had read it.

Yes I used them to optimize it before, but I did not realize that after changing a few things using an SVG editor It reset them.
I will make the changes. Thank you.

jmb:
I've taken a look and it is not in the correct place as it doesn't fit in the green circle area as per the guideline.svg

image

Once you have adjusted the first path to be in the right place, then giving it an `id="a"` and doing the `use` that I mentioned will help reduce the file size. I recommend using https://yqnn.github.io/svg-path-editor/ to play with the rounding and optimise the path - you only need to paste in the d value from a path element.

Also, the layout of the file still doesn't match the guidelines given (it even mentions that in the PR template right at the top here!)

PR #1082, opened 2025-10-30
Update Amazon S3 icon

empeig:

* [X] I have read [the contributing guidelines](https://github.com/edent/SuperTinyIcons/blob/master/CONTRIBUTING.md)
* [X] I have run `python check.py` and received a ✅
```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```

### DIFF
--- a/images/svg/amazon_s3.svg
+++ b/images/svg/amazon_s3.svg
@@ -2,4 +2,4 @@
 aria-label="Amazon Simple Storage Service" role="img"
 viewBox="0 0 512 512"><path
 d="m0 0H512V512H0"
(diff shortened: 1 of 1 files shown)


Eiim:
You can definitely reduce the file size by reducing the precision. Rounding to one decimal point is plenty fine, and you might be able to do 0 in places. I also wonder if it wouldn't be easier and smaller to incorporate the ellipse in the bucket path? It can be replicated with two arc commands.

empeig:
Thank you for the feedback. I'll try to optimize a bit

romain-dartigues:
You also can shrink a few bytes by  using properties without the style; ie:
`stroke-width="12" stroke="#fff" fill="none"` instead of `style="stroke-width:12px;stroke:#fff;fill:none"` → saves 4

About the arcs commands empeig mentioned, you should give a try to the excellent https://yqnn.github.io/svg-path-editor/

empeig (commented):
I took into account all your suggestions:

- replaced the ellipse by two arcs. It is difficult to have the same shape as the original which uses 4 arcs but I think it is close enough. See this file for reference (red is the original svg, white the SuperTiny svg shape):
EditV2

- Rounding to one decimal

- Removed the style from properties

PR #1084, opened 2025-11-01
Add LastFm

carmoran0:

```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```


jmb (changes requested):
This is looking good and fits within the green circle area of the guidelines. I think you can remove the group and rectangle background that's within it (set the main background path as the red colour) and should apply the transform to the path (if you're using Inkscape, removing the transform from the XML editor should apply the transform, or you can use https://yqnn.github.io/svg-path-editor/ ). There is also an opacity which shouldn't be there.

PR #1100, opened 2026-02-22
Add Slackware Linux

romain-dartigues:
As there is no formal version of the logo:

* I took a capital "S" of Courier Bold © 1991 Digital Typeface Corp. (see POLM citation below)
* Used the http://www.slackware.com/favicon.ico as source

> Q: What is the original Slackware font as used in the old-style Slackware-with-an-underline style logo ?
>
> A: I've checked the main Slackware logo (name-with-underline) against some fonts,
> and it looks like it's Courier, not Courier New as you have it in the FAQ.
> […]
> Courier New is inappropriate because it rounds off the ends of the letters, which isn't very slack of it.
>
> — Paul O'Leary McCann http://www.slackware.com/~msimons/slackware/grfx/grfxfaq.txt

<details><summary>Thank you for your contribution!  Before sending this Pull Request, please confirm the following:</summary>

```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```

</details>


edent:
I'm not sure. The S looks rather different from the one at https://arm.slackware.com/images/slackware.png

Looks like it needs to be squashed and the ends flatter.

romain-dartigues:
Oh, you are not bothered by the distances between the elements (^_^').
Yeah, I don't know either, the "official" description is explicit *against* the rounded version, but, as you posted, it is the most used version.

In a recent post from AlienBob, we can see the old and the modern logos:
https://blog.slackware.nl/a-slackware-theme-for-your-grub/

Let me know which version you prefer:
1. Following the original rules, using Courier slackware
2. New version, in a circle 
slackware
3. New version, on black (as in the bootloader); I would personally prefer 
slackware

PR #1104, opened 2026-05-06
Bambu Lab

meyerweb:

* [ ] I have run `python check.py` and received a ✅
```svg
aria-label="..." role="img"
viewBox="0 0 512 512"><path
d="m0 0H512V512H0"
fill="#fff"/> ... </svg>
```


meyerweb:
I have this set up as a draft because I think everything is okay, but when I run `python check.py` I get:

```
Traceback (most recent call last):
  File "/Users/emeyer/repos/SuperTinyIcons/check.py", line 66, in <module>
    raise RuntimeError("Validator did not start in time")
RuntimeError: Validator did not start in time
```

meyerweb:
Also, I can’t find any actual official brand guidelines for Bambu.  Not sure if that’s a blocker or not.

PR #1106, opened 2026-05-11
Bambulab

meyerweb:
Update to the Bambu Lab SVG, reducing it by 36 bytes (~10%).  Also fixing its Aria label, which was slightly inaccurate.


meyerweb:
Sorry for the accreting commits.  Git and I only barely get along.

